PR #15, opened 2013-06-29
interrupt opcode handling

aarzilli:


navytux:
Update: commit https://github.com/hanwen/go-fuse/commit/e028a29e09 added INTERRUPT request handling by propagating it to closing `req.cancel` channel. However that cancel chan is not yet available in any way to user request handlers.

PR #38, opened 2015-04-19
Retry on EINTR for read/write syscalls

akalin:
Currently, it treats it like any other error, which seems like a bug.

### DIFF
--- a/fuse/server.go
+++ b/fuse/server.go
@@ -198,6 +198,25 @@ func (ms *Server) DebugData() string {
 // What is a good number?  Maybe the number of CPUs?
 const _MAX_READERS = 2
 
+// Retries the given function until it doesn't return syscall.EINTR.
+// This is similar to the HANDLE_EINTR() macro from Chromium ( see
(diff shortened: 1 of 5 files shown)


akalin:
I signed it!

hanwen:
The code looks OK. Can you squash the commits together and write some more comment under which circumstances this might trigger? Does this fix the following log line?

2015/04/21 10:46:03 Unimplemented opcode INTERRUPT
2015/04/21 10:46:03 writer: Write/Writev failed, err: 2=no such file or directory. opcode: INTERRUPT

if your handleEINTR is the same is libc's temp_failure_retry, allude to that and maybe rename it.

akalin:
Okay, I squashed the commits and wrote a comment for handleEINTR. I decided against the temp_failure_retry name, as I think handleEINTR is more explicit.

I put the description of the triggering cases (running under a debugger) in the commit description, as I felt it fit better there. I can also put it in the function comment, too, if you want.

I don't think it'll fix that log line -- fuse has a separate INTERRUPT opcode, which is different as I understand it.

Please take another look!

hanwen:
I had to fix up 2 details, but pushed it now. Thanks!

PR #40, opened 2015-04-20
Make BufferPool.AllocBuffer() return a buffer of the correct size

### LINKED ISSUE #39: bufferPoolImpl.AllocBuffer() returns a too-big buffer
The buffer it returns has length a multiple of the page size, which can be larger than the input size.

This can cause problems, because len(buf) is usually used instead of the input size. This manifests itself in the following way:

- fuse does, say, a READDIR request with size 1500 bytes.
- `doReadDir` in opcode.go gets a buffer of size 4096 and uses it.
- server.filesystem.ReadDir() happily fills up the buffer up to 4096 bytes.
- we pass the buffer back to fuse, which barfs on OS X. (Linux seems to be more permissive about this.)

This doesn't seem to be a problem with files, likely because READ requests already have sizes of 4096.

Pull request coming soon.

akalin:
Also add tests.

This fixes #39.

### DIFF
--- a/fuse/bufferpool.go
+++ b/fuse/bufferpool.go
@@ -78,7 +78,7 @@ func (p *bufferPoolImpl) AllocBuffer(size uint32) []byte {
 	pages := sz / PAGESIZE
 
 	b := p.getPool(pages).Get().([]byte)
-	return b
+	return b[:size]
 }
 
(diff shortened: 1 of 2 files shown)


hanwen:
I've squashed and merged. Thanks!

PR #42, opened 2015-04-22
Use generation numbers properly

akalin:
Move the generation field from nodefs/FileSystemConnector to handleMap.
Have handleMap.Register() return both the inode number and the generation number.
Increment the generation number whenever Register() reuses a previously-used node ID.
Add test.

### DIFF
--- a/fuse/nodefs/fsconnector.go
+++ b/fuse/nodefs/fsconnector.go
@@ -8,7 +8,6 @@ import (
 	"log"
 	"path/filepath"
 	"strings"
-	"sync/atomic"
 	"time"
 	"unsafe"
 
@@ -22,10 +21,6 @@ var paranoia = false
 // structs of uint32/uint64) to operations on Go objects representing
 // files and directories.
 type FileSystemConnector struct {
-	// Used as the generation inodes. This must be 64-bit aligned,
-	// for sync/atomic on i386 to work properly.
-	generation uint64
-
 	debug bool
 
 	// Callbacks for talking back to the kernel.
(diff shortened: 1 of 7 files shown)


hanwen:
see here for more background on NFS https://web.archive.org/web/20100109210802/http://old.nabble.com/fuse_entry_param-td26661841.html

hanwen:
Overall, this looks reasonable, but your original bug had a detailed scenario of things could go wrong on OSX. Can you destill that into a testcase, verify it is broken before these changes and fixed afterwards?

akalin:
Ah, okay, I didn't know that (node, generation) had to be unique for the whole lifetime of the filesystem.

In that case, I'll initialize the generation number back to time.Now(), which seems reasonable.

However, it looks like inodes/handles are already 64-bit-sized. In that case, is there even a need to keep track of the free IDs? We could just have a uint64 counter and always increment it for call to Register() that needs a new inode number. Then the generation number can be effectively constant (= timestamp of when fuse was started).

I'll also try to write a testcase for it...although, I think the existing ones are failing for me so I'll need to fix those first.

hanwen:
You could do that, but you'd have to use a hash map for the lookups which
is somewhat more expensive.

You don't have to fix all of the tests, but please add a case for this so
we don't regress.

Op wo 22 apr. 2015 23:17 schreef Fred Akalin <notifications@github.com>:

akalin:
Ah, I see...forgot about having to use a hash map!

Unfortunately, it looks like the existing nodefs tests don't really work on OS X, so I'd have to fix those first before copying one of them.

akalin:
Oh, I didn't get around to putting back time.Now() as the generation number, plus tests! But I can do that in another CL I guess.

PR #53, opened 2015-09-19
Fix issues uncovered by xfstests

rfjakob:
Adds the `FsName` member to `MountOptions` and uses it in the loopback example

Before:
    
    Filesystem                 Type                   1K-blocks      Used  Available Use% Mounted on
    pathfs.pathInode           fuse.pathfs.pathInode    4015252   3219628     795624  81% /tmp/mountpoint
    
After:
    
    Filesystem                 Type           1K-blocks      Used  Available Use% Mounted on
    /tmp/original              fuse.loopback    4015252   3219620     795632  81% /tmp/mountpoint

Also, I enabled ClientInodes. This is in preparation of getting xfstests to run on loopbackfs.

### DIFF
--- a/fuse/mount_linux.go
+++ b/fuse/mount_linux.go
@@ -1,6 +1,7 @@
 package fuse
 
 import (
+	"path"
 	"bytes"
 	"fmt"
 	"log"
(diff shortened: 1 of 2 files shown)


rfjakob:
I signed it!

hanwen:
thanks for all these changes. I will try to look soon, but am rather busy.

I think we probably don't want to set ClientInodes as a default, because it has subtle semantics. LoopbackFS uses path names as identifier, which is fundamentally at odds with posix if you look at the details (files with hardlinks and such.)

rfjakob:
Hi there, I have addressed your comments and rebased to master.
I have also dropped the ClientInodes patch for now. I will add a flag to enable it at run-time next week.

rfjakob:
Already merged, closing the pull request.

Issue #58, opened 2015-09-28
Fails xfstests generic/258, setting times of 'testfile.txt': Invalid argument


rfjakob:
Loopbackfs fails xfstests generic/258 because timestamps before 1970 are not handled correctly.

    touch -t 196001010101 testfile.txt
    touch: setting times of 'testfile.txt': Invalid argument

I have tracked this down to https://github.com/hanwen/go-fuse/blob/master/fuse/nodefs/files_linux.go#L35 . The Futimes returns the EINVAL, but UnixNano() seems to be the culprit. I have printed out the values around L35:

    a = 1960-01-01 01:01:00 +0100 CET, m = 1960-01-01 01:01:00 +0100 CET
    n = -315619140000000000
    tv = [{-315619139 -999999} {-315619139 -999999}]

rfjakob:
I think the problem is the negative value in `tv.Usec`.

This is a bug in go core: https://github.com/golang/go/blob/master/src/syscall/syscall_linux_amd64.go#L99

rfjakob:
Fixed by 065c5910cd24bfbc385e0b4384d04229602b955c, closing the ticket.

PR #59, opened 2015-09-29
loopback: Replace buggy syscall.NsecToTimeval()

rfjakob:
syscall.NsecToTimespec does not work properly for times
before 1970, see https://github.com/golang/go/issues/12777

This caused xfstests generic/258 to fail.

Fixes issue #58.

Additionally, this fixes a bug in line 46 where `a` was used instead of `m`

### DIFF
--- a/fuse/nodefs/files.go
+++ b/fuse/nodefs/files.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/hanwen/go-fuse/fuse"
 )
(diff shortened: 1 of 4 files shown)


rfjakob:
Indeed. Also, needs more refactoring. The function in files_darwin.go is identical. Will fix that up.

rfjakob:
Ok, done.

hanwen:
thanks, merged.

for style, note this commit though:
https://github.com/hanwen/go-fuse/commit/c2130ac9e52cdd7ec9b819922fc9a8c5aba9d8dc

Issue #60, opened 2015-09-29
Fails xfstests generic/221, UTIME_OMIT does not work


rfjakob:
xfstests generic/221 runs [t_futimens.c](https://github.com/rfjakob/fuse-xfstests/blob/gocryptfs/src/t_futimens.c), that executes the following syscall:

    utimensat(3, NULL, {{1000000000, 0}, UTIME_OMIT}, 0) = -1 EINVAL (Invalid argument)

As you can see, it fails with EINVAL. Stracing the loopback process shows:

    [pid 27783] utimes("/proc/self/fd/3", {{1000000000, 0}, {0, 1073741822}} <unfinished ...>
    [pid 27783] <... utimes resumed> )      = -1 EINVAL (Invalid argument)

Note that 1073741822 == UTIME_OMIT. However, this special value only works for `utimens` and `utimensat` that take nanoseconds, not for `utimes` that only takes microseconds!

*** This is the last issue with the "generic" xfstests ***

Failing tests are now:

* generic/002 (missing hardlink support)
* generic/035 (known bug https://github.com/hanwen/go-fuse/issues/55)
* generic/089 (missing hardlink support)
* generic/221 (this bug)
* generic/236 (missing hardlink support)

Everything else passes or is skipped.

PR #61, opened 2015-10-04
loopback: Use UtimesNano instead of Futimes

rfjakob:
Futimes() only takes microsecond resolution while the FUSE call
Utimens() wants nanosecond precision.

This is why UTIME_OMIT did not work - this change fixes issue #60.


rfjakob:
The Futimens syscall is not implemented! Check out https://golang.org/pkg/syscall/

hanwen:
yes, that's what I am saying. See fuse/syscall_linux.go

rfjakob:
Wow, that looks scary. I will take a look!

hanwen:
it is simpler than it looks on first glance; you'll figure it out :-)

I think the Go library code may /proc for compatibility with older kernels.

rfjakob:
Interestingly, `futimens` is not a syscall. It is just a glibc function that wraps `utimensat`.

So I have implemented Futimens() just like glibc does it and now use that function in the Utimens() FUSE call.

rfjakob:
There is UtimesNanoAt ( https://godoc.org/golang.org/x/sys/unix#UtimesNanoAt ) which looked promising for a moment, but it does not allow to pass a nil path.

rfjakob:
Ok, I have moved `futimens` into nodefs/syscall.go and made it private.

Issue #62, opened 2015-10-07
loopback: Missing "." and ".." entries


rfjakob:
```
0 jakob@brikett:~/src/github.com/hanwen/go-fuse/example/loopback$ ls -la backing
total 8
drwxr-xr-x. 2 jakob jakob 4096 26. Aug 00:38 .
drwxr-xr-x. 5 jakob jakob 4096  5. Okt 23:21 ..
0 jakob@brikett:~/src/github.com/hanwen/go-fuse/example/loopback$ ls -la mountpoint
total 0
```

rfjakob:
Seems to be a regression in 8783b8e89957fdfdcf61b5a7472cd3d5a99b3e0b. Reverting it brings the entries back.

PR #63, opened 2015-10-09
Get "." and ".." back

rfjakob:
Also adds a test for them.

The change passed an xfstests run.

Fixes issue #62

### DIFF
--- a/fuse/nodefs/dir.go
+++ b/fuse/nodefs/dir.go
@@ -70,10 +70,6 @@ func (d *connectorDir) ReadDirPlus(input *fuse.ReadIn, out *fuse.DirEntryList) (
 			continue
 		}
 
-		if e.Name == "." || e.Name == ".." {
-			continue
-		}
-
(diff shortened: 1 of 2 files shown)


fuse/nodefs/dir.go
```diff
@@ -70,10 +70,6 @@ func (d *connectorDir) ReadDirPlus(input *fuse.ReadIn, out *fuse.DirEntryList) (
 			continue
 		}
 
-		if e.Name == "." || e.Name == ".." {
-			continue
-		}
-
```
hanwen: you should skip the lookup for . and .. below

rfjakob: Shouldn't readdirplus also return the extra infos for "." and ".."?

rfjakob:
Ok, the kernel does not seem to ever stat "." and "..", instead it stats the (parent) dir directly. Fixed.

rfjakob:
Already merged, thanks! Closing the pull request.

PR #68, opened 2015-10-20
nodefs: do not overwrite the "out.Ino" inode number if it is already set

rfjakob:
"out.Ino" is the user-facing inode number that is displayed by "ls -li".

This change allows loopback filesystems to expose the inode number of
the underlying filesystem.
As a "real" inode number, this one is stable for the lifetime of the file,
even across remounts.

libfuse has the `use_ino` option that allows the inode numbers to be set arbitrarily as well. This has been used by EncFS for years.

The Linux kernel stores the inode number in `orig_ino`. Grepping in `fs/fuse` shows that it is only read once in `fuse_update_attributes`, and only if the `stat` argument is not NULL. There is only one caller that passes that argument: `fuse_getattr`.

In short, the kernel does not use the inode number for anything but reporting it via `stat()`, hence setting the inode number is safe.

This functionality will be used for improving log output in gocryptfs, where file names cannot be used because they are encrypted.

### DIFF
--- a/fuse/nodefs/fsconnector.go
+++ b/fuse/nodefs/fsconnector.go
@@ -89,7 +89,9 @@ func (c *rawBridge) childLookup(out *fuse.EntryOut, n *Inode, context *fuse.Cont
 	n.Node().GetAttr((*fuse.Attr)(&out.Attr), nil, context)
 	n.mount.fillEntry(out)
 	out.NodeId, out.Generation = c.fsConn().lookupUpdate(n)
-	out.Ino = out.NodeId
+	if out.Ino == 0 {
+		out.Ino = out.NodeId
+	}
 	if out.Nlink == 0 {
 		// With Nlink == 0, newer kernels will refuse link
 		// operations.
--- a/fuse/nodefs/fsmount.go
+++ b/fuse/nodefs/fsmount.go
@@ -70,7 +70,9 @@ func (m *fileSystemMount) fillEntry(out *fuse.EntryOut) {
 func (m *fileSystemMount) fillAttr(out *fuse.AttrOut, nodeId uint64) {
 	splitDuration(m.options.AttrTimeout, &out.AttrValid, &out.AttrValidNsec)
 	m.setOwner(&out.Attr)
-	out.Ino = nodeId
+	if out.Ino == 0 {
+		out.Ino = nodeId
+	}
 }
 
 func (m *fileSystemMount) getOpenedFile(h uint64) *openedFile {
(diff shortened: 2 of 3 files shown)


rfjakob:
This change passes both `go check ./...` (that is, it does not introduce new test failures) and fuse-xfstests.

hanwen:
are you double-plus sure that we can never get a NodeId from kernel that uses this user-supplied inode number? Maybe ask on fuse-devel@ ?

rfjakob:
Uhmm, pretty much, yes? Libfuse has an option to allow it  (quoting from `man mount.fuse`):

```
       use_ino
              Honor  the  st_ino  field  in  kernel functions getattr() and
              fill_dir(). This value is used to fill in the st_ino field in
              the stat(2), lstat(2), fstat(2) functions and the d_ino field
              in the readdir(2) function. The filesystem does not  have  to
              guarantee  uniqueness, however some applications rely on this
              value being unique for the whole filesystem.
```

...and EncFS uses this option to expose the inode numbers.

If this option is not set, libfuse overwrites the inode number with the nodeid, just like go-fuse currently does - check out https://github.com/rfjakob/fuse/blob/master/lib/fuse.c#L1452 .

rfjakob:
I mean, EncFS is pretty mature, is this good enough evidence?

hanwen:
"The filesystem does not  have  to
              guarantee  uniqueness, "


I guess this means that inode number isn't used for anything really.


On Tue, Oct 20, 2015 at 9:14 PM, rfjakob <notifications@github.com> wrote:

> I mean, EncFS is pretty mature, is this good enough evidence?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hanwen/go-fuse/pull/68#issuecomment-149672890>.
>



-- 
Han-Wen Nienhuys - hanwenn@gmail.com - http://www.xs4all.nl/~hanwen

rfjakob:
The kernel stores the inode number in `orig_ino`. Grepping in `fs/fuse` shows that it is only read once in `fuse_update_attributes`, and only if the `stat` argument is not NULL. There is only one caller that passes that argument: `fuse_getattr`.

Should be safe.

hanwen:
Can you update your PR with this information? Seems valuable to keep around.

On Tue, Oct 20, 2015 at 10:23 PM, rfjakob <notifications@github.com> wrote:

> The kernel stores the inode number in orig_ino. Grepping in fs/fuse shows
> that it is only read once in fuse_update_attributes, and only if the stat
> argument is not NULL. There is only one caller that passes that argument:
> fuse_getattr.
>
> Should be safe.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hanwen/go-fuse/pull/68#issuecomment-149691570>.
>



-- 
Han-Wen Nienhuys - hanwenn@gmail.com - http://www.xs4all.nl/~hanwen

rfjakob:
Merging via the github webinterface forces a merge commit that contains the pull request text.

However, I just noticed you don't do that, so I have updated the commit message with the full text.

PR #77, opened 2015-11-16
Resolved os x problem

y-okubo:


rfjakob:
Pull Req https://github.com/hanwen/go-fuse/pull/89 should fix this properly.

PR #79, opened 2015-11-26
Fix crash on listxattr syscall with NULL list argument.

Edgeworth:
Steps to reproduce:
1. Run a loopback filesystem
2. Run getfattr on a file with no extended attributes.


Edgeworth:
Done~

hanwen:
committed as 850263fac3ab5077c157b3da6ccd78011b27a147

PR #80, opened 2015-11-26
Add fsync handling to FileSystemConnector.

Edgeworth:
I'm not entirely sure if this is the correct way to do this. Calls to Read and Write get shunted through Node before finally reaching File.

### DIFF
--- a/fuse/nodefs/fsops.go
+++ b/fuse/nodefs/fsops.go
@@ -21,6 +21,13 @@ func (c *FileSystemConnector) RawFS() fuse.RawFileSystem {
 type rawBridge FileSystemConnector
 
 func (c *rawBridge) Fsync(input *fuse.FsyncIn) fuse.Status {
+	node := c.toInode(input.NodeId)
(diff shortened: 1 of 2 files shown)


Edgeworth:
Done~

hanwen:
Your test doesn't actually check that the sync is received by the FS.

On Mon, Dec 7, 2015 at 12:46 AM, Eliot Courtney <notifications@github.com>
wrote:

> Done~
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hanwen/go-fuse/pull/80#issuecomment-162372770>.
>



-- 
Han-Wen Nienhuys - hanwenn@gmail.com - http://www.xs4all.nl/~hanwen

Edgeworth:
Sorry about that -- added.

While adding the FsyncCalled field, I noticed there was an old GetAttrCalled flag that is not used anymore (was originally used to see if GetAttr was called, but got partially removed along the way), by the way.

hanwen:
148d3830fabe1501ac2d8cabcd7d5fc0005c8a33

Issue #81, opened 2015-11-28
loopbackfs: Utimens on symlinks does not work


rfjakob:
Reproducer:
```
$ ./loopback /tmp/b /tmp/a &
Mounted!
$ cd /tmp/b
$ ln -s /nonexisting/file symlink
$ touch --no-dereference symlink
touch: setting times of ‘symlink’: No such file or directory
```
The problem is that Utimens ( https://github.com/hanwen/go-fuse/blob/master/fuse/pathfs/loopback.go#L121 ) operates on the target of the symlink and not on the symlink itself, like it should.
This also happens when extracting a tar that has a symlink in it, and this is how I noticed it.

Utimens has to use `utimensat` with `AT_SYMLINK_NOFOLLOW`, but the Go API has no way to do that, so we have to define our own syscall wrapper like in https://github.com/hanwen/go-fuse/commit/fa19882f2f985f994f7f2019618252f2e745bdbe .

Will send a patch tomorrow.

PR #82, opened 2015-11-29
loopback: fix Utimens on symlinks

rfjakob:
The problem was that os.Chtimes() operates on the target of the
symlink and not on the symlink itself.
When tar extracts an archive containing a symlink, it actually wants
to set the times on the symlink. This can also be tested using
"touch -h".

This patch adds a wrapper for the Linux utimensat(2) syscall
that allows passing flags along with a unit test.

It uses the flag AT_SYMLINK_NOFOLLOW to implement
loopbackFileSystem.Utimens() properly.

It also add handling for UTIME_NOW and UTIME_OMIT that was already
present in loopbackFile.Utimens() but missing in loopbackFileSystem.

Fixes issue 81 ( https://github.com/hanwen/go-fuse/issues/81 ).

### DIFF
--- a/fuse/nodefs/files.go
+++ b/fuse/nodefs/files.go
@@ -204,6 +204,7 @@ func (f *loopbackFile) GetAttr(a *fuse.Attr) fuse.Status {
 const _UTIME_NOW = ((1 << 30) - 1)
 const _UTIME_OMIT = ((1 << 30) - 2)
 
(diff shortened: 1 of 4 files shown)


rfjakob:
Already merged as https://github.com/hanwen/go-fuse/commit/d36ee9dd90398a0ab40a3c26a0ccb21dfdaaf96a , closing, thanks

PR #86, opened 2015-12-16
Run go fmt ./...

nnnlife:


hanwen:
thanks for the heads up. I've formatted the code and pushed.

(I wish github could enforce this serverside)

PR #88, opened 2015-12-23
fix compilation error on ubuntu-32bit

nnnlife:
time.Unix() returns int64 regardless of target platform but field of timeval is
defined as int32 values on ubuntu-32bit
NsecToTimeval has a nanosecond problem(negative value) before epoch but still can use a second field


hanwen:
I had some other patches in these files; could you rebase if this is still relevant?

Issue #71, opened 2015-11-03
all.bash doesn't succeed on OS X - el capitan


varung:
```
bash all.bash 
+ for target in '"clean"' '"install"'
+ for d in fuse fuse/pathfs fuse/test zipfs unionfs example/hello example/loopback example/zipfs example/multizip example/unionfs example/memfs example/autounionfs
+ test clean = install
+ go clean github.com/hanwen/go-fuse/fuse
+ for d in fuse fuse/pathfs fuse/test zipfs unionfs example/hello example/loopback example/zipfs example/multizip example/unionfs example/memfs example/autounionfs
+ test clean = install
+ go clean github.com/hanwen/go-fuse/fuse/pathfs
+ for d in fuse fuse/pathfs fuse/test zipfs unionfs example/hello example/loopback example/zipfs example/multizip example/unionfs example/memfs example/autounionfs
+ test clean = install
+ go clean github.com/hanwen/go-fuse/fuse/test
+ for d in fuse fuse/pathfs fuse/test zipfs unionfs example/hello example/loopback example/zipfs example/multizip example/unionfs example/memfs example/autounionfs
+ test clean = install
+ go clean github.com/hanwen/go-fuse/zipfs
+ for d in fuse fuse/pathfs fuse/test zipfs unionfs example/hello example/loopback example/zipfs example/multizip example/unionfs example/memfs example/autounionfs
+ test clean = install
+ go clean github.com/hanwen/go-fuse/unionfs
+ for d in fuse fuse/pathfs fuse/test zipfs unionfs example/hello example/loopback example/zipfs example/multizip example/unionfs example/memfs example/autounionfs
+ test clean = install
+ go clean github.com/hanwen/go-fuse/example/hello
+ for d in fuse fuse/pathfs fuse/test zipfs unionfs example/hello example/loopback example/zipfs example/multizip example/unionfs example/memfs example/autounionfs
+ test clean = install
+ go clean github.com/hanwen/go-fuse/example/loopback
+ for d in fuse fuse/pathfs fuse/test zipfs unionfs example/hello example/loopback example/zipfs example/multizip example/unionfs example/memfs example/autounionfs
+ test clean = install
+ go clean github.com/hanwen/go-fuse/example/zipfs
+ for d in fuse fuse/pathfs fuse/test zipfs unionfs example/hello example/loopback example/zipfs example/multizip example/unionfs example/memfs example/autounionfs
+ test clean = install
[...]

PR #87, opened 2015-12-19
partial fix: fix build on osx, osx test failures remain

glycerine:
While this fixes the compilation issues on OSX, it should be reviewed for correctness--no attempt to understand the full semantics of the nodefs and pathfs subprojects was attempted.

There are still test failures (on OSX) -- show below. But at least we can build to evaluation and test further.

~~~
jaten@Jasons-MacBook-Pro:~/go/src/github.com/hanwen/go-fuse$ ./all.bash
+ for target in '"clean"' '"install"'
+ for d in fuse fuse/pathfs fuse/test zipfs unionfs example/hello example/loopback example/zipfs example/multizip example/unionfs example/m\
emfs example/autounionfs

...

+ go install github.com/hanwen/go-fuse/example/autounionfs
+ for d in fuse zipfs unionfs
+ cd fuse
+ go test github.com/hanwen/go-fuse/fuse
ok      github.com/hanwen/go-fuse/fuse  0.008s
+ go test -race github.com/hanwen/go-fuse/fuse
ok      github.com/hanwen/go-fuse/fuse  1.013s
+ for d in fuse zipfs unionfs
+ cd zipfs
+ go test github.com/hanwen/go-fuse/zipfs
--- FAIL: TestMultiZipReadonly (0.04s)
    multizip_test.go:49: Must fail writing in root.
--- FAIL: TestMultiZipFs (0.05s)
    multizip_test.go:69: wrong names return. []
    multizip_test.go:74: Symlink failed: symlink /Users/jaten/go/src/github.com/hanwen/go-fuse/zipfs/test.zip /var/folders/2x/hm9gp5ys3k9gm\
m5f_vzm_6wc0000gn/T/846065809/config/zipmount: no such file or directory
--- FAIL: TestZipFs (0.04s)
    zipfs_test.go:50: wrong length []
    zipfs_test.go:54: Stat failed: stat /var/folders/2x/hm9gp5ys3k9gmm5f_vzm_6wc0000gn/T/611642556/subdir: no such file or directory
--- FAIL: TestLinkCount (0.04s)
    zipfs_test.go:90: Stat failed: stat /var/folders/2x/hm9gp5ys3k9gmm5f_vzm_6wc0000gn/T/946217707/file.txt: no such file or directory
FAIL
FAIL    github.com/hanwen/go-fuse/zipfs 0.180s
jaten@Jasons-MacBook-Pro:~/go/src/github.com/hanwen/go-fuse$
~~~

Issue #78, opened 2015-11-18
build error on mac 


vChrysanthemum:
# github.com/hanwen/go-fuse/fuse/nodefs
../github.com/hanwen/go-fuse/fuse/nodefs/files_darwin.go:5: imported and not used: "time"
../github.com/hanwen/go-fuse/fuse/nodefs/syscall.go:11: undefined: syscall.SYS_UTIMENSAT

alexmojaki:
+1

The problematic callis:

    syscall.Syscall6(syscall.SYS_UTIMENSAT, uintptr(fd), 0, uintptr(unsafe.Pointer(times)), uintptr(0), 0, 0)

I'm assuming that this is the file being used:

https://golang.org/src/syscall/syscall_bsd.go

`SYS_UTIMENSAT` indeed isn't defined but there seems to be a function that can help at line 509:

```
   509	func UtimesNano(path string, ts []Timespec) error {
   510		// TODO: The BSDs can do utimensat with SYS_UTIMENSAT but it
   511		// isn't supported by darwin so this uses utimes instead
```

PR #89, opened 2016-01-09
Fix several Utimens-related build problems on OSX

rfjakob:
Properly implementing Utimens caused the OSX build to break because
the utimensat system call is not available there.

This patch reverts the OSX implementation of Utimens to use
syscall.Utimes(). This is not actually correct but the best we can do.

The Linux implementation is unchanged.

Thanks to @spaghetti2514 who reported the issue and tested the patches. With these changes, he was able to build gocryptfs on OSX.


rfjakob:
Fixes issues #71, #78

rfjakob:
Obsoletes #77, #87

hanwen:
thanks!

PR #97, opened 2016-01-31
Have pathInode.Truncate and friends use the file handle, if available

### LINKED ISSUE #95: pathInode.Truncate and friends ignore `file` argument
For example, https://github.com/hanwen/go-fuse/blob/master/fuse/pathfs/pathfs.go#L666 . The passed `file` argument is ignored and instead it uses a random (?) file handle.

This sometimes picks a file handle that is already closed when [gocryptfs](https://nuetzlich.net/gocryptfs/) handles the truncate request, leading to unneccessary error messages and a general sense of "how the heck can this even happen" at yours sincerly.

Is there a reason for not using `file` or are you ok if I fix this?

rfjakob:
If the file handle is specified we should use it instead of
picking one at random.

Fixes #95.

### DIFF
--- a/all.bash
+++ b/all.bash
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -eux
+set -eu
 
 for target in "clean" "install" ; do
   for d in fuse fuse/pathfs fuse/test zipfs unionfs \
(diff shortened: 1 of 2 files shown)


rfjakob:
Tests added. Turns out only for fallocate and ftruncate the kernel actually passes in a file handle. For fstat, fchmod and fchown, the added code for checking if the file handle is there is just for future-proofing.

hanwen:
this adds API surface, which I'd like to avoid, and using a global variable at that.

Can't you just do 
 
  mount loopback
  f = open("mnt/a")
  mv("orig/a", "orig/b")
  touch("orig/a")
  
 ftruncate(f)

check that b is changed, not a.

rfjakob:
That test would fall short i believe, because the kernel always identifies
files by node id. That's not a proof that the file handle was used.

But maybe deleting the file works, will check.

hanwen:
but in PathFS, the node ID is transformed back into a path, so you can tell if the FD is used by looking at which file was touched.

rfjakob:
Yes, but it tries first to use an open fh

rfjakob:
I have not found a way to test this without adding API surface. So to move this forward I have dropped the test and updated the patch. After all, implementing the test had quite a benefit for my understanding of what happens, which I wrote down in the code comments.

I have also added a patch that enables running the integration tests in `fuse/test`, and also fixed the bug that they uncovered immediately.

rfjakob:
You have already fixed the mtime bug in ac2190345467cc7bb600f33be326c68af12b8bc0, dropped the patch and rebased to latest master.

PR #98, opened 2016-02-08
Fix "Found linked inode, but Nlink == 1" bug

rfjakob:
Apart from adding the missing write into clientInodeMap, I have
added a few comments about what happens. Hopefully helps the
next person looking at this code.


hanwen:
was merged as ba787feaaba0ee14ee3585cb70c8e96ae0ae8a64

Issue #99, opened 2016-02-12
Fails xfstests generic/013, renames files to ".deleted"


rfjakob:
I don't know why I haven't seen this earlier in my gocryptfs testing, but the hard link tracking (clientInodes) in pathfs seems to have a few issues (test case below). pathfs seems to lose track of linked files, internally renaming them to ".deleted". These files show up in directory listing but cannot be deleted nor stat'ed. Sometimes it crashes with a null pointer dereference in pathfs.

I am working on fixing the issues. It is not ready for merging yet, but it's probably going be a bigger patch. If you already want to take a look, the state right now with lots of added debugging code is here: https://github.com/rfjakob/go-fuse/commit/fd910e744f38180a9456a2086681ed3e5b60de52

Test case, with gocryptfs mounted on /tmp/b:
```
( set -e;
  while true; do
  ~/src/fuse-xfstests/ltp/fsstress -p 1 -z -f rmdir=10 -f link=10 -f creat=10 -f mkdir=10 -f rename=30 -f stat=30 -f unlink=30 -f truncate=20 -m 8 -n 1000 -d /tmp/b
  rm -R /tmp/b/*
  sleep 1
done )
```

hanwen:
If something shows up in directory listings, then you must have somehow got these into the OpenDir() result.

can you post the stack trace for the NPE ?

rfjakob:
Sorry, my description was unclear:

The files show up under their normal name, but gocryptfs gets passed
".deleted" on any operation, hence stat and everything else fails.

Will post the NPE sunday evening!

rfjakob:
I cannot reproduce the NPE with a fresh checkout of go-fuse. Must have been caused by my debug code.

On the plus side, I can reproduce this with the loopback example (with ClientInodes enabled, patch: https://github.com/rfjakob/go-fuse/commit/1f50fb4dd13aa46eb282271662b6166c42417ed9 , otherwise no changes)

```
$ ( set -e; while true; do ~/src/fuse-xfstests/ltp/fsstress -p 1 -z -f rmdir=10 -f link=10 -f creat=10 -f mkdir=10 -f rename=30 -f stat=30 -f unlink=30 -f truncate=20 -m 8 -n 1000 -d /tmp/b; rm -R /tmp/b/*; sleep 1; done )
seed = 1456229227
[...]

PR #102, opened 2016-02-27
loopback example: add "-l" flag for enabling hard links

rfjakob:
Enabling ClientInodes in the loopback example means that hard links can be tested easier using external tools like fsstress (see #99).


rfjakob:
Updated the pull request to add a command line flag, `-l`, instead of hard-coding ClientInodes = true.

PR #103, opened 2016-03-17
Correctness fixes for passing xfstests generic/013 (fsstress)

rfjakob:
As promised, this is the patch series to fix the xfstests generic/013 failure (issue #99).

It strengthens the consistency checking in the clientInodes code and fixes several issues and race conditions in pathfs and nodefs.

The code now passes thousands of fsstress iterations, where it would fail after tens of iterations before.

### DIFF
--- a/fuse/nodefs/fsconnector.go
+++ b/fuse/nodefs/fsconnector.go
@@ -114,6 +114,7 @@ func (c *FileSystemConnector) lookupUpdate(node *Inode) (id, generation uint64) 
 	return
 }
 
+// forgetUpdate - decrement reference counter for "nodeID" by "forgetCount".
 // Must run outside treeLock.
 func (c *FileSystemConnector) forgetUpdate(nodeID uint64, forgetCount int) {
 	if nodeID == fuse.FUSE_ROOT_ID {
@@ -140,7 +141,6 @@ func (c *FileSystemConnector) InodeHandleCount() int {
 }
 
 // Must hold treeLock.
-
 func (c *FileSystemConnector) recursiveConsiderDropInode(n *Inode) (drop bool) {
 	delChildren := []string{}
 	for k, v := range n.children {
--- a/fuse/nodefs/fsconnector.go
+++ b/fuse/nodefs/fsconnector.go
@@ -121,6 +121,7 @@ func (c *FileSystemConnector) registerNode(node *Inode) (id, generation uint64) 
 // forgetUpdate - decrement reference counter for "nodeID" by "forgetCount".
 // Must run outside treeLock.
 func (c *FileSystemConnector) forgetUpdate(nodeID uint64, forgetCount int) {
+
 	c.forgetLock.Lock()
 	defer c.forgetLock.Unlock()
 
@@ -134,12 +135,18 @@ func (c *FileSystemConnector) forgetUpdate(nodeID uint64, forgetCount int) {
 
 	if forgotten, handled := c.inodeMap.Forget(nodeID, forgetCount); forgotten {
 		node := (*Inode)(unsafe.Pointer(handled))
+		if node.IsDir() && len(node.Children()) > 0 {
+			// This directory node has children that have not yet been forgotten.
+			// We don't want to remove it from the tree (i.e. remove it from its parent)
+			// because the children may have parent pointers to it.
+			// A later LOOKUP will return this node with a new NodeID.
+			return
+		}
 		node.mount.treeLock.Lock()
 		// Remove Inode from all parents
 		// We have to create a new slice first because rmChild modifies node.parents
 		nParents := []*Inode{}
 		nParents = append(nParents, node.parents...)
-		fmt.Printf("forgetUpdate: removing from %d parents\n", len(nParents))
 		for _, p := range(nParents) {
 			p.rmChildByRef(node)
 		}
--- a/fuse/nodefs/fsconnector.go
+++ b/fuse/nodefs/fsconnector.go
@@ -59,7 +59,7 @@ func NewFileSystemConnector(root Node, opts *Options) (c *FileSystemConnector) {
 
 	// FUSE does not issue a LOOKUP for 1 (obviously), but it does
 	// issue a forget.  This lookupUpdate is to make the counts match.
-	c.lookupUpdate(c.rootNode)
+	c.registerNode(c.rootNode)
 
 	return c
 }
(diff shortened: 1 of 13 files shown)


hanwen:
Wow! Please give me a couple of days to digest to this.

Op do 17 mrt. 2016 23:11 schreef rfjakob <notifications@github.com>:

fuse/nodefs/fsconnector.go
```diff
@@ -126,9 +135,23 @@ func (c *FileSystemConnector) forgetUpdate(nodeID uint64, forgetCount int) {
...
+		if node.IsDir() && len(node.Children()) > 0 {
+			// This directory node has children that have not yet been forgotten.
+			// We don't want to remove it from the tree (i.e. remove it from its parent)
+			// because the children may have parent pointers to it.
+			// A later LOOKUP will return this node with a new NodeID.
+			return
+		}
 		node.mount.treeLock.Lock()
-		c.recursiveConsiderDropInode(node)
+		// Remove Inode from all parents
```
hanwen: I don't agree with this: this conflates two things, which is: 1) the inode being known to the kernel, and 2) whether or not the the inode should be removed from the tree.  1) is a mattter of erasing the ID from the map, and 2) is a type of garbage collection. 

The clientInodes support is completely in pathfs right now, and I don't see why the support for it has to migrate into nodefs. Can you verify that libfuse needs to modify the lowlevel support code to support hardlinks? If not, we should not need to either.

All of the comments you added to existing code seem fine to me. If you can prepare a separate commit that just adds those, I can merge that without any issues.

hanwen: in particular, the new code now doesn't call Deletable which seems wrong.

rfjakob: Hi there, thank you for starting the review so quick!

I don't think is only about garbage collection, this is also about
(1) one file should never be represented by multiple nodes in go space
(2) after onforget() has been called, the object should never be used again

As far as i can see, we have to cleanly remove the node from the tree to
achieve that. And we need parent pointers for that.

(Will check the ondeleted issue later!)

rfjakob: Thinking about it, probably nodefs should not beware of the tree structure at all? What do you think?

hanwen: If it's not aware of tree structure, then what does it understand? The
lookup operation maps quite naturally to a nice containing a string indexed
map for its children.

Op di 22 mrt. 2016 22:32 schreef rfjakob <notifications@github.com>:

hanwen: PS. I'll be out for a few days this week.

Op di 22 mrt. 2016 23:21 schreef Han-Wen Nienhuys <hanwenn@gmail.com>:

rfjakob: Only the nodeid mapping. Like libfuse with its "fuse lowlevel" interface.

hanwen: that sounds pretty difficult to use. 

hte whole point of nodefs is that many FS already construct a tree in memory (see eg. gitfs or zipfs). With the current setup you can construct the entire tree ahead of time as inodes, and the lookup/forget just return node IDs as needed. This is also why you can't destruct the inode on a forget message.

rfjakob: That's true and it's a nice feature. The problem for pathfs is this:
https://github.com/hanwen/go-fuse/blob/master/fuse/nodefs/fsops.go#L72 . Nodefs decides which node to use for a lookup, and does so based just on the name. This is not good enough for hard link tracking UNLESS the tree in nodefs is perfectly in sync with the underlying file system.

hanwen: Yes, but you can make the tree be in perfect sync if the FUSE daemon is the
thing acting on the file system, right?

rfjakob: It looks like pathfs could just stop using the children map that nodefs provides. The lookup already calls into the filesystem if it does not know the child, so this would need no code changes in nodefs.

This would get rid of the duplicate tree and allow the hard link tracking to live exclusively in pathfs.

rfjakob: Keeping it in sync: Yes, that is what I did with this patch series. I thought you did not like it?

hanwen: The backlinks needed for inodes are tied to the nodefs.Node implementation,
so they already are in the same structure, at least before your patch
series.

rfjakob: If we have the full tree in nodefs (actually, with hard links, it's not a tree anymore, it's DAG I believe), then pathfs should drop it's own Parent pointers to avoid duplication.

rfjakob: Ah ok, so the "parents" map should live in pathfs? This would work I guess.

hanwen: but forget is not a filesystem operation, so in itself a forget should have
no effect on the tree. It is just a garbage collection that the kernel does
occasionally.

the tree is a DAG but in a limited sense. Directories cannot have
hardlinks, for example.

nodefs does not have parent pointers, because there is no need for them.
You only need them if you want to implement hardlinks on top of a
path-based API.

rfjakob: So I see three options:
1) Have the full DAG in nodefs (= add parent pointers).
2) Have the "down" direction (children) of the DAG in nodefs, the "up" direction (parents) in pathfs. Looks a bit strange but should work.
3) Have the full DAG in pathfs (= add children and parent pointers)

From what I understand you favor (2) ?

hanwen: From what I know it already looks like 2.

Op wo 23 mrt. 2016 10:16 schreef rfjakob <notifications@github.com>:

rfjakob: Yes, there already is a "Parent" pointer - this needs to become a map.

I will rework the patch series and start a new pull request once I am ready!

hanwen: Great. I didn't do a map at the time because I thought it would be
expensive, but TBH I did not measure it. If it is an issue, maybe you can
have a single parent pointer in case of directory nodes, eg.

  // use in case this is a directory
  dirParent *pathNode
  // use for non-directory nodes.
  parents  map[*pathNode]struct{}

rfjakob:
Obsoleted by v2 of the patch series, https://github.com/hanwen/go-fuse/pull/104

PR #104, opened 2016-03-27
Correctness fixes for passing xfstests generic/013 (fsstress), v2

rfjakob:
v2 of the patch series consolidates the changes into fewer patches and skips some intrusive unionfs changes. The clientinode handling is not factored out into its own file but integrated more tightly into pathfs, together with the new "parents" map. This may come at a performance cost for everybody not using hard links, but the returns is a simpler code base. My canonical gocryptfs benchmark (hard links enabled) showed no measurable changes whatsover.
As with v1, loopbackfs not passes thousands of iterations.

### DIFF
--- a/fuse/nodefs/fsops.go
+++ b/fuse/nodefs/fsops.go
@@ -68,7 +68,11 @@ func (c *FileSystemConnector) lookupMountUpdate(out *fuse.Attr, mount *fileSyste
 	return mount.mountInode, fuse.OK
 }
 
+// internalLookup - execute a lookup without affecting NodeId reference counts
 func (c *FileSystemConnector) internalLookup(out *fuse.Attr, parent *Inode, name string, header *fuse.InHeader) (node *Inode, code fuse.Status) {
+
+	// We may already know the child because it was created using Create or Mkdir
+	// or from an earlier lookup. The kernel submits new lookups periodically.
 	child := parent.GetChild(name)
 	if child != nil && child.mountPoint != nil {
 		return c.lookupMountUpdate(out, child.mountPoint)
--- a/fuse/nodefs/fsops.go
+++ b/fuse/nodefs/fsops.go
@@ -70,6 +70,8 @@ func (c *FileSystemConnector) lookupMountUpdate(out *fuse.Attr, mount *fileSyste
 
 // internalLookup - execute a lookup without affecting NodeId reference counts
 func (c *FileSystemConnector) internalLookup(out *fuse.Attr, parent *Inode, name string, header *fuse.InHeader) (node *Inode, code fuse.Status) {
+	c.forgetMu.RLock()
+	defer c.forgetMu.RUnlock()
 
 	// We may already know the child because it was created using Create or Mkdir
 	// or from an earlier lookup. The kernel submits new lookups periodically.
--- a/fuse/nodefs/inode.go
+++ b/fuse/nodefs/inode.go
@@ -157,6 +157,7 @@ func (n *Inode) RmChild(name string) (ch *Inode) {
 //////////////////////////////////////////////////////////////
 // private
 
+// addChild - Add "child" to our children under name "name".
 // Must be called with treeLock for the mount held.
 func (n *Inode) addChild(name string, child *Inode) {
 	if paranoia {
@@ -168,6 +169,7 @@ func (n *Inode) addChild(name string, child *Inode) {
 	n.children[name] = child
 }
 
+// rmChild - Drop "name" from our children
 // Must be called with treeLock for the mount held.
 func (n *Inode) rmChild(name string) (ch *Inode) {
 	ch = n.children[name]
--- a/fuse/nodefs/inode.go
+++ b/fuse/nodefs/inode.go
@@ -113,7 +113,7 @@ func (n *Inode) IsDir() bool {
 	return n.children != nil
 }
 
-// NewChild adds a new child inode to this inode.
+// NewChild creates an Inode, sets that Inode using fsi.SetInode() and calls AddChild.
 func (n *Inode) NewChild(name string, isDir bool, fsi Node) *Inode {
 	ch := newInode(isDir, fsi)
 	ch.mount = n.mount
@@ -131,11 +131,14 @@ func (n *Inode) GetChild(name string) (child *Inode) {
 	return child
 }
 
-// AddChild adds a child inode. The parent inode must be a directory
-// node.
+// AddChild adds an existing "*Inode" as a child to ourself, under the name "name".
+// Panics for non-directory Inodes.
 func (n *Inode) AddChild(name string, child *Inode) {
 	if child == nil {
 		log.Panicf("adding nil child as %q", name)
+	}
+	if !n.IsDir() {
+		log.Panicf("adding child %q to non-directory", name)
 	}
 	n.mount.treeLock.Lock()
 	n.addChild(name, child)
--- a/fuse/pathfs/pathfs.go
+++ b/fuse/pathfs/pathfs.go
@@ -6,18 +6,12 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/hanwen/go-fuse/fuse"
 	"github.com/hanwen/go-fuse/fuse/nodefs"
 )
-
-// A parent pointer: node should be reachable as parent.children[name]
-type clientInodePath struct {
-	parent *pathInode
-	name   string
-	node   *pathInode
-}
 
 // PathNodeFs is the file system that can translate an inode back to a
 // path.  The path name is then used to call into an object that has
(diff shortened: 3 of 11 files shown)


fuse/server.go
```diff
@@ -28,6 +28,21 @@ type Server struct {
...
+	//                     LOOKUP "foo"
+	//                     --> return NodeId 2
+	// --> forget 2
+	//
+	// At this point the kernel holds a reference to NodeId 2 that has
+	// been forgotten inside go-fuse.
+	forgetMu sync.RWMutex
+
 	// I/O with kernel and daemon.
 	mountFd int
```
hanwen: do you have any proof that this is a problem?

forget and lookup have reference counts. In your scenario, the count for  "foo" starts at 1, goes to 2 (lookup "foo") and then back to 1 (forget). 

The kernel will have forgotten node 2, but will know about it once the Lookup result is processed.

fuse/nodefs/inode.go
```diff
@@ -131,12 +131,15 @@ func (n *Inode) GetChild(name string) (child *Inode) {
 	return child
 }
 
-// AddChild adds a child inode. The parent inode must be a directory
-// node.
+// AddChild adds an existing "*Inode" as a child to ourself, under the name "name".
+// Panics for non-directory Inodes.
```
hanwen: s/panic.../the node N must be a directory Inode./

fuse/pathfs/pathfs.go
```diff
@@ -210,16 +198,23 @@ func (fs *PathNodeFs) Root() nodefs.Node {
...
 type pathInode struct {
 	pathFs *PathNodeFs
 	fs     FileSystem
-	Name   string
 
-	// This is nil at the root of the mount.
-	Parent *pathInode
+	// Due to hard links, a file can have many parents and one ore more names
+	// under each parent. The map is indexed by (name, parent) tuples, the
+	// value is unused.
```
hanwen: use struct{} for increased memory efficicency

fuse/pathfs/pathfs.go
```diff
@@ -210,16 +198,23 @@ func (fs *PathNodeFs) Root() nodefs.Node {
 	return fs.root
 }
 
+type nameAndParent struct {
+	name   string
+	parent *pathInode
```
hanwen: this structure implies an invariant. Do you have a verification method that will check the invariant paranoically?

fuse/pathfs/pathfs.go
```diff
@@ -255,147 +242,100 @@ func (n *pathInode) SetInode(node *nodefs.Inode) {
...
-// Reread all client nodes below this node.  Must run outside the treeLock.
-func (n *pathInode) updateClientInodes() {
-	n.GetAttr(&fuse.Attr{}, nil, nil)
-	for _, ch := range n.Inode().FsChildren() {
-		ch.Node().(*pathInode).updateClientInodes()
+// getParent - returns a random parent of this pathInode and the name under this
+// parent.
+// The caller must hold pathLock.
+func (n *pathInode) getParent() (*pathInode, string) {
+	for k, _ := range n.parents {
```
hanwen: add a TODO somewhere: the parents map is only necessary for non-directory nodes. We could shortcut some indirections by having a parent pointer for directory nodes. (should benchmark first, though.)

fuse/pathfs/pathfs.go
```diff
@@ -255,147 +242,100 @@ func (n *pathInode) SetInode(node *nodefs.Inode) {
...
+		log.Printf("getPath: non-root node without parents, returning dummyPath=%q", dummyPath)
+		return dummyPath
 	}
-	pathLen--
 
-	if p != p.pathFs.root {
-		n.pathFs.pathLock.RUnlock()
-		return ".deleted"
+	// The simple solution is to collect names, and reverse join
+	// them. This is a hot path and may need optimization to avoid
```
hanwen: remove "may". It is a hot path.

fuse/pathfs/pathfs.go
```diff
@@ -255,147 +242,100 @@ func (n *pathInode) SetInode(node *nodefs.Inode) {
...
+
+	// Walk up to the root and collect names in pathSegments
+	var pathSegments []string
+	var nextName string
+	nextParent := n
+	for len(nextParent.parents) > 0 {
+		nextParent, nextName = nextParent.getParent()
+		pathSegments = append(pathSegments, nextName)
+
+		if len(pathSegments) > syscall.PathMax {
```
hanwen: is this an attempt at avoiding infinite loops? I  think it's OK for paths to be longer than PathMax. After all, these paths don't go into syscalls

fuse/pathfs/pathfs.go
```diff
@@ -255,147 +242,100 @@ func (n *pathInode) SetInode(node *nodefs.Inode) {
...
-		l := len(p.Name)
-		copy(pathBytes[end-l:], p.Name)
-		end -= len(p.Name) + 1
-		if end > 0 {
-			pathBytes[end] = '/'
+	// Join segments to full path
+	fullPath := pathSegments[0]
+	if len(pathSegments) > 1 {
+		for _, segment := range pathSegments[1:] {
+			fullPath = segment + "/" + fullPath
```
hanwen: do something smarter. This is quadratic due to realloc of fullpath.

fuse/nodefs/fsops.go
```diff
@@ -68,7 +68,11 @@ func (c *FileSystemConnector) lookupMountUpdate(out *fuse.Attr, mount *fileSyste
 	return mount.mountInode, fuse.OK
 }
 
+// internalLookup - execute a lookup without affecting NodeId reference counts
 func (c *FileSystemConnector) internalLookup(out *fuse.Attr, parent *Inode, name string, header *fuse.InHeader) (node *Inode, code fuse.Status) {
+
+	// We may already know the child because it was created using Create or Mkdir
+	// or from an earlier lookup. The kernel submits new lookups periodically.
```
hanwen: Also, you can create the tree ahead of time.

fuse/nodefs/inode.go
```diff
@@ -168,6 +169,7 @@ func (n *Inode) addChild(name string, child *Inode) {
 	n.children[name] = child
 }
 
+// rmChild - Drop "name" from our children
```
hanwen: add .

rfjakob:
Yes the explanation for `forgetMu` was wrong. I have pushed the lock down into nodefs where the race really is: https://github.com/hanwen/go-fuse/pull/104/commits/35a0b1edaadaf4c8078b939e7c87a0edb1cee7de
I have verified that it still passes fsstress (and it immediately fails with the lock removed).

fuse/pathfs/pathfs.go
rfjakob: Yes it is, it helped during development, but if your filesystem does not
call into the kernel there is no real reason to limit the length. Will drop
the check.

fuse/nodefs/fsops.go
```diff
@@ -70,6 +70,8 @@ func (c *FileSystemConnector) lookupMountUpdate(out *fuse.Attr, mount *fileSyste
 
 // internalLookup - execute a lookup without affecting NodeId reference counts
 func (c *FileSystemConnector) internalLookup(out *fuse.Attr, parent *Inode, name string, header *fuse.InHeader) (node *Inode, code fuse.Status) {
+	c.forgetMu.RLock()
```
hanwen: this is really problematic. You're holding a global FS lock while getattr executes. GetAttr is a user specified function, so it often does expensive or blocking actions (eg. network lookups.), so this lock would effectively kill performance. The fact that fsstress passes with this lock also doesn't say much, because the lock may mask some other race condition.

hanwen: In the original code, OnForget is only called after it has successfully removed from the tree with parent.rmChild(). So, the kernel can never get a new reference to the forgotten child. Of course, that doesn't stop the pathfs from getting its parent mappings all screwed up. Maybe you can have an OnAdd() and OnRemove() which run inside the nodefs treeLock?

rfjakob: No, the original code did not remove the forgotten node at all from the tree. I could not because it does not have parent pointers. It removed all the children, but not the node itself. By the way it also never called OnForget on the node itself.

rfjakob: Oh but protecting `child := parent.GetChild(name)` ( fsops.go:72 ) with the lock should be enough. That way GetAttr can run outside the lock. Will check if it still passes.

rfjakob: Oh well. I think we additionally need a per-inode forgetMu.
But first, please take a look at https://github.com/rfjakob/go-fuse/commit/ab4f2be05c683b6d8090b3b19106be9e401e488b and tell me if you think nulling fsInode is correct here. I think it is, but it exposes lots of bugs. I can trigger NPEs in a few iterations of fsstress.

Issue #107, opened 2016-05-16
nodefs API: cannot implement chgrp


rfjakob:
chgrp(1) actually calls chown(2) with `owner` set to -1.

Right now, the nodefs code does not handle chgrp at all, but that could be fixed easily. The real problem is that the parameters for `Node.Chown()` are defined as uint32 instead of int as for `syscall.Chown()`

How to fix this? Pass `0xFFFFFFFF` instead of -1 ?

rfjakob:
This should do it. If you agree I will submit it together with a test case for loopbackfs.

```
diff --git a/fuse/nodefs/fsops.go b/fuse/nodefs/fsops.go
index 459b75c..4c3120d 100644
--- a/fuse/nodefs/fsops.go
+++ b/fuse/nodefs/fsops.go
@@ -201,7 +201,15 @@ func (c *rawBridge) SetAttr(input *fuse.SetAttrIn, out *fuse.AttrOut) (code fuse
                code = node.fsInode.Chmod(f, permissions, &input.Context)
        }
        if code.Ok() && (input.Valid&(fuse.FATTR_UID|fuse.FATTR_GID) != 0) {
-               code = node.fsInode.Chown(f, uint32(input.Uid), uint32(input.Gid), &input.Context)
+               var uid uint32 = ^uint32(0) // means "don't care" in chown(2)
+               var gid uint32 = ^uint32(0)
+               if input.Valid&fuse.FATTR_UID != 0 {
+                       uid = input.Uid
+               }
+               if input.Valid&fuse.FATTR_GID != 0 {
+                       gid = input.Gid
+               }
+               code = node.fsInode.Chown(f, uid, gid, &input.Context)
        }
        if code.Ok() && input.Valid&fuse.FATTR_SIZE != 0 {
                code = node.fsInode.Truncate(f, input.Size, &input.Context)
```

PR #108, opened 2016-05-17
nodefs: fix chgrp

rfjakob:
Fixes https://github.com/hanwen/go-fuse/issues/107 .

Uses `^uint32(0)` as a uint32 substitute for `-1`.


rfjakob:
I did not push this to Gerrit because it's a simple fix and I hoped we can fast-track this.

hanwen:
github is mirrored from gerrithub nowadays, so the change has to go to gerrit first. I uploaded your change and merged it.

rfjakob:
Didn't know that, thanks a lot!

PR #112, opened 2016-07-14
loopback: use AT_FDCWD in utimensat call

rfjakob:
When mounting using relative paths, "touch" on directories fails:

  $ loopback b a &
  $ cd b
  $ mkdir foo
  $ touch foo
  touch: setting times of 'foo': Not a directory

strace:

  [pid 30185] utimensat(0, "a/foo", [{1468441847, 0}, {1468441847, 0}], AT_SYMLINK_NOFOLLOW <unfinished ...>
  [...]
  [pid 30185] <... utimensat resumed> )   = -1 ENOTDIR (Not a directory)

Change-Id: If78f67fd8cfee946bf1a1112b81dd7fe959afe2b

### DIFF
--- a/fuse/pathfs/loopback.go
+++ b/fuse/pathfs/loopback.go
@@ -25,6 +25,12 @@ type loopbackFileSystem struct {
 // system.  Its main purpose is to provide test coverage without
 // having to build a synthetic filesystem.
 func NewLoopbackFileSystem(root string) FileSystem {
+	// Make sure the Root path is absolute to avoid problems when the
+	// application changes working directory.
+	root, err := filepath.Abs(root)
+	if err != nil {
+		panic(err)
+	}
 	return &loopbackFileSystem{
 		FileSystem: NewDefaultFileSystem(),
 		Root:       root,


hanwen:
why does the touch fail? If the arg to loopback must be absolute, it should be doc'd as a requirement.

rfjakob:
I looked at it again, the actual bug was elsewhere, pull request updated.
Fix is https://github.com/rfjakob/go-fuse/commit/94de61d1095fb6dc23d777ace05f96f258b3b03e

hanwen:
ok. - this all sounds too complicated. Lets deref the argument on construction. Painful bit is that we'd have to change the signature of the NewLoopbackFS() function.

rfjakob:
No, that's not good enough - how do you fix this problem in Open? ( https://github.com/hanwen/go-fuse/blob/master/fuse/pathfs/loopback.go#L99 )

hanwen:
sorry for letting this linger. I don't understand. 

My suggestion is to have NewLoopbackFS() just try to absolutize the path. 

what is the problem you see with Open() ?

rfjakob:
Oh, misunderstanding on my side!
Making the path absolute in NewLoopbackFilesystem sounds good. Will update the PR.

rfjakob:
Updated the patch.

PR #116, opened 2016-08-02
Add test for #115

ahl:
Let me know if you had something different in mind. I confirmed that this fails prior to my patch.

### DIFF
--- a/fuse/test/mount_test.go
+++ b/fuse/test/mount_test.go
@@ -4,6 +4,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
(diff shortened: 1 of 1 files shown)


hanwen:
looks almost like what I was thinking of, but I'd rather see a test of a public interface, ie. trying to access the mountpoint (eg os.Lstat), and veriifying that that works. You'd probably have to move the test into test/ to get acces to a working filesystem

ahl:
I tried a different approach.

hanwen:
yeah, that's more like it.
thanks!

hanwen:
cherry picked as d5da28da3917f2d902497a36eca71d9e68183189

PR #117, opened 2016-08-04
fixed a bunch of tests on darwin

ahl:
Everything except for unionfs is now passing. Still chasing down some bugs there, but wanted to see if there was interest before continuing.

Found a legit bug in unionfs:
When the Finder sent it an unexpected getxattr unionfs responded ENOSYS because the subordinate filesystem returned it. Returning ENOSYS causes FUSE to *never* send more getxattrs. We can intercept that and replace it with ENODATA.

### DIFF
--- a/README
+++ b/README
@@ -104,8 +104,7 @@ go-fuse works somewhat on OSX. Known limitations:
 * OSX has trouble with concurrent reads from the FUSE device, leading
   to performance concerns.
 
-* A bunch of tests under fuse/test/ fail, and the functionality they
-  test for OSX is likely broken.
+* Tests are expected to pass; report any failure as a bug!
 
 
 CREDITS
--- a/fuse/server.go
+++ b/fuse/server.go
@@ -61,14 +61,6 @@ func (ms *Server) SetDebug(dbg bool) {
 	ms.opts.Debug = dbg
 }
 
-func (ms *Server) SupportsNotify(op int) bool {
-	switch op {
-	case NOTIFY_INVAL_ENTRY:
-		return ms.kernelSettings.Major >= 7 && ms.kernelSettings.Minor >= 12
-	}
-	return false
-}
-
 // KernelSettings returns the Init message from the kernel, so
 // filesystems can adapt to availability of features of the kernel
 // driver.
@@ -560,13 +552,13 @@ func (ms *Server) EntryNotify(parent uint64, name string) Status {
 
 // SupportsVersion returns true if the kernel supports the given
 // protocol version or newer.
-func (in *InitIn) SupportsVersion(maj, min uint32) bool {
+func (in InitIn) SupportsVersion(maj, min uint32) bool {
 	return in.Major >= maj && in.Minor >= min
 }
 
 // SupportsNotify returns whether a certain notification type is
 // supported. Pass any of the NOTIFY_INVAL_* types as argument.
-func (in *InitIn) SupportsNotify(notifyType int) bool {
+func (in InitIn) SupportsNotify(notifyType int) bool {
 	switch notifyType {
 	case NOTIFY_INVAL_ENTRY:
 		return in.SupportsVersion(7, 12)
--- a/fuse/server.go
+++ b/fuse/server.go
@@ -25,9 +25,6 @@ const (
 // Server contains the logic for reading from the FUSE device and
 // translating it to RawFileSystem interface calls.
 type Server struct {
-	versionMajor uint32
-	versionMinor uint32
-
 	// Empty if unmounted.
 	mountPoint string
 	fileSystem RawFileSystem
@@ -64,8 +61,12 @@ func (ms *Server) SetDebug(dbg bool) {
 	ms.opts.Debug = dbg
 }
 
-func (ms *Server) Version() (major, minor uint32) {
-	return ms.versionMinor, ms.versionMinor
+func (ms *Server) SupportsNotify(op int) bool {
+	switch op {
+	case NOTIFY_INVAL_ENTRY:
+		return ms.kernelSettings.Major >= 7 && ms.kernelSettings.Minor >= 12
+	}
+	return false
 }
 
 // KernelSettings returns the Init message from the kernel, so
--- a/fuse/server.go
+++ b/fuse/server.go
@@ -25,6 +25,9 @@ const (
 // Server contains the logic for reading from the FUSE device and
 // translating it to RawFileSystem interface calls.
 type Server struct {
+	versionMajor uint32
+	versionMinor uint32
+
 	// Empty if unmounted.
 	mountPoint string
 	fileSystem RawFileSystem
@@ -59,6 +62,10 @@ type Server struct {
 func (ms *Server) SetDebug(dbg bool) {
 	// This will typically trigger the race detector.
 	ms.opts.Debug = dbg
+}
+
+func (ms *Server) Version() (major, minor uint32) {
+	return ms.versionMinor, ms.versionMinor
 }
 
 // KernelSettings returns the Init message from the kernel, so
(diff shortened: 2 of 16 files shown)


fuse/server.go
```diff
@@ -25,6 +25,9 @@ const (
 // Server contains the logic for reading from the FUSE device and
 // translating it to RawFileSystem interface calls.
 type Server struct {
+	versionMajor uint32
```
hanwen: this is already available in the kernelSettings member.

fuse/test/loopback_darwin_test.go
```diff
@@ -10,9 +10,14 @@ import (
...
 	s.Fsid = empty.Fsid
-	// s.Spare = empty.Spare
-	// TODO - figure out what this is for.
-	s.Flags = 0
+	s.Owner = empty.Owner
+	s.Flags = empty.Flags
+	s.Fssubtype = empty.Fssubtype
+	s.Fstypename = empty.Fstypename
+	s.Mntonname = empty.Mntonname
+	s.Mntfromname = empty.Mntfromname
```
hanwen: maybe it's easier to init to empty and copy the fields that FUSE can set?

unionfs/autounion_test.go
```diff
@@ -142,14 +144,22 @@ func TestAutoFsSymlink(t *testing.T) {
...
-	fi, _ = os.Lstat(wd + "/mnt/manual1")
-	if fi != nil {
-		t.Error("Should not have file:", fi)
-	}
-
-	_, err = ioutil.ReadDir(wd + "/mnt/config")
-	if err != nil {
-		t.Fatalf("ReadDir failed: %v", err)
+	// Note that FUSE API 7.11 introduced the notification support required to cause the node to be immediately removed. This check will fail in earlier versions.
+	if _, minor := server.Version(); minor >= 11 {
```
hanwen: it's unlikely that major will change from 7 , but maybe easier to add 

  func (in *InitIn) SupportsOpcode(opcode int)

and/or  

  func (in *InitIn) SupportsFeature(flag int)

unionfs/unionfs.go
```diff
@@ -712,7 +712,12 @@ func (fs *unionFS) GetXAttr(name string, attr string, context *fuse.Context) ([]
 	}
 	r := fs.getBranch(name)
 	if r.branch >= 0 {
-		return fs.fileSystems[r.branch].GetXAttr(name, attr, context)
+		value, code := fs.fileSystems[r.branch].GetXAttr(name, attr, context)
+		// We intercept ENOSYS and replace it with ENODATA; returning ENOSYS would cause FUSE to never to GetXAttr() again.
+		if code == fuse.ENOSYS {
+			return nil, fuse.ENODATA
+		}
+		return value, code
```
hanwen: this can be a separate CL. ENOSYS should be intercepted in opcode.go (there should be handling for ENOSYS in connection with the DisableXattr option.)

README
```diff
@@ -104,8 +104,7 @@ go-fuse works somewhat on OSX. Known limitations:
 * OSX has trouble with concurrent reads from the FUSE device, leading
```
hanwen: thanks for looking into this. It's much appreciated!

unionfs/autounion_test.go
ahl: Good idea. I'd suggest that it belongs in Server rather than InitIn

fuse/test/loopback_darwin_test.go
ahl: Good idea.

unionfs/unionfs.go
ahl: I think this makes more sense in unionfs. Typically a filesystem will either implement GetXAttr or they'll return ENOSYS in which case the existing behavior is desirable. I don't think you want to force users to know they should set DisableXAttr to avoid a bunch of repeated useless requests.

Can you imagine a scenario where what you're describing would be beneficial? This issue would seem to be one with unionfs not one that applies more generally.

unionfs/autounion_test.go
hanwen: It's all keyed on the InitIn data, though, so I think it should rather be on InitIn.

unionfs/unionfs.go
hanwen: Support xattr is a property of the entire mount point, since FUSE can't decide which parts of the filesystem do or do not support it. Hence the support should also be where the mountpoint is handled.

This is already partly in place, see 7b281480b21a59f54a0c86451e7de13cd9be736f.   - it just needs to transform ENOSYS into ENODATA there.

unionfs/autounion_test.go
ahl: so... turn `kernelSettings` into `KernelSettings`?

unionfs/unionfs.go
ahl: It seems like either ENOSYS should be converted into ENODATA in opcode.go or that the approach in https://github.com/hanwen/go-fuse/commit/7b281480b21a59f54a0c86451e7de13cd9be736f should be continued (i.e. change ENOSYS to ENODATA in fuse/pathfs/default.go).

If we choose the former, then I'd suggest reverting the previous changes in fuse/defaultraw.go and fuse/nodefs/defaultnode.go.

Let me know.

ahl:
pushed an updated diff for review.

fuse/server.go
```diff
@@ -64,8 +61,12 @@ func (ms *Server) SetDebug(dbg bool) {
 	ms.opts.Debug = dbg
 }
 
-func (ms *Server) Version() (major, minor uint32) {
-	return ms.versionMinor, ms.versionMinor
+func (ms *Server) SupportsNotify(op int) bool {
+	switch op {
+	case NOTIFY_INVAL_ENTRY:
+		return ms.kernelSettings.Major >= 7 && ms.kernelSettings.Minor >= 12
```
ahl: confirming the version when support was added:
https://github.com/osxfuse/kext/blob/5639d3e9c9aa32c532a0faef562d6c459a485f8d/osxfuse/fuse_ipc.h#L511

unionfs/unionfs.go
hanwen: I think opcode.go is the best place. It guarantees that this stops biting us.

fuse/server.go
hanwen: Let me provide something for query what the server supports.

(to answer another question: there is already a KernelSEttings() method.)

hanwen: see https://github.com/hanwen/go-fuse/commit/3cf08068e8f6ff0c7961023efdf318b44a3ba719

ahl: Seems needlessly indirect, but, fine, I guess I'll use it now.

unionfs/unionfs.go
ahl: I'd note that that seems like the wrong methodology and would potentially create surprises for users, but fine. Shall I revert the incomplete work in https://github.com/hanwen/go-fuse/commit/7b281480b21a59f54a0c86451e7de13cd9be736f and change those ENODATA returns to ENOSYS?

ahl:
updated to use your new interface.

ahl:
all tests are now passing on darwin; finally!

fuse/server.go
```diff
@@ -560,13 +552,13 @@ func (ms *Server) EntryNotify(parent uint64, name string) Status {
 
 // SupportsVersion returns true if the kernel supports the given
 // protocol version or newer.
-func (in *InitIn) SupportsVersion(maj, min uint32) bool {
+func (in InitIn) SupportsVersion(maj, min uint32) bool {
```
hanwen: why? this is making a copy of InitIn now.

hanwen:
I've squashed this and cherry-picked (except the InitIn change). Thanks!

fuse/server.go
ahl: Because `KernelSettings()` returns a non-pointer. The zipfs and unionfs tests don't build right now on Darwin; do they build and work on Linux?

hanwen: ah, you're right. I've changed KernelSettings to return a pointer.

ahl: I didn't take that approach because I figured the goal was to give the user something that wouldn't mutate the original.

PR #118, opened 2016-08-08
added Link() to FileSystem

codegamc:
Wont build without type FileSystem having a link().  pathfs/locking.go
fails to build because it expects type FileSystem to have method
.link(), so method was added to the interface


codegamc:
I signed it!

fuse/pathfs/api.go
```diff
@@ -71,7 +71,7 @@ type FileSystem interface {
 	// Symlinks.
 	Symlink(value string, linkName string, context *fuse.Context) (code fuse.Status)
 	Readlink(name string, context *fuse.Context) (string, fuse.Status)
-
+	Link(oldName string, newName string, context *fuse.Context) (code fuse.Status)
```
hanwen: I see link in l.46.

PR #119, opened 2016-08-09
fix ENOATTR handling for darwin

ahl:
On Mac OS X the Finder loses its mind if GetXAttr returns ENODATA rather than ENOATTR. On Linux, ENOATTR is just an alias for ENOATTR.

Note that the Linux man page references ENOATTR:
http://man7.org/linux/man-pages/man2/fgetxattr.2.html

You can also see the discussion regarding the golang syscall package:
https://github.com/golang/go/issues/753

Both discussion note that ENOATTR is a synonym for ENODATA.

On Mac OS X, that's not the case:

```
...
#if __DARWIN_C_LEVEL >= __DARWIN_C_FULL
#define ENOATTR         93              /* Attribute not found */
#endif

#define EBADMSG         94              /* Bad message */
#define EMULTIHOP       95              /* Reserved */
#define ENODATA         96              /* No message available on STREAM */
...
```

Note that we could take the approach of removing ENODATA altogether, but I worry that that would break existing users. Unfortunately Go doesn't seem to have a mechanism to deprecate an interface.

Still another approach would be to translate ENODATA to ENOATTR on darwin (since they're actually different codes), but that seems like a kludge.

### DIFF
--- a/fuse/defaultraw.go
+++ b/fuse/defaultraw.go
@@ -86,7 +86,7 @@ func (fs *defaultRawFileSystem) GetXAttrSize(header *InHeader, attr string) (siz
 }
 
 func (fs *defaultRawFileSystem) GetXAttrData(header *InHeader, attr string) (data []byte, code Status) {
-	return nil, ENODATA
+	return nil, ENOATTR
 }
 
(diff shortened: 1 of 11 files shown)


hanwen:
thanks, cherry-picked.

ahl:
Thanks for fixing the syscall.ENOATTR on Linux

PR #121, opened 2016-08-09
loopback: re-enable nanoseconds for dates after 1970

rfjakob:
For dates before 1970, syscall.NsecToTimespec() incorrectly
returns negative nanoseconds.
Ticket: https://github.com/golang/go/issues/12777

Handle that case by zeroing out Nsec if it is negative,
but otherwise keep the value.

This makes nanoseconds work for the vast majority of use
cases.


rfjakob:
I'll re-submit this in a simple form.

Issue #132, opened 2016-09-26
Why does not loopbackFile.Read write the result into buf?


chmduquesne:
I am referring to this line:

https://github.com/hanwen/go-fuse/blob/master/fuse/nodefs/files.go#L128

It seems like apart from its size, the input buffer is ignored and the result of the read is directly sent through the fuse wire protocol. Does it mean that buf is useless?

Is there a way to rewrite this function so that it make use of buf?

hanwen:
for data coming out of a file, it's less work to leave the data in the kernel and use splice.

You could just use f.Read() on the passed-in buf.

PR #133, opened 2016-09-30
Loopback: Making the code for Read() consistent with the code for Write()

chmduquesne:
In issue #132, I was wondering why the read buffer was not filled in the Read() function. Now that I understand better what is going on, I would like to suggest this patch, which IMHO makes the code more understandable.

I think this version is simpler for the newcomer, because one can see how the result is filled directly from the Read() function, rather than having to dive in the code of readResultFd.

Feel free to take the patch, unless you believe this does not improve the readability.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: go
+sudo: required
+dist: trusty
+before_install:
+  - sudo apt-get install -qq fuse
+  - sudo modprobe fuse
+  - sudo /bin/sh -c 'echo user_allow_other > /etc/fuse.conf'
+  - sudo chmod 666 /dev/fuse
+  - sudo chown root:$USER /etc/fuse.conf
+script:
+  - go get -t -v ./...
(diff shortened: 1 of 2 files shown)


chmduquesne:
I signed this a while ago!

Back then I did not register a github username with the CLA.

Also my email is set in my git commits, but I don't want to use my
gmail address for that.

chmduquesne:
Hi,

I am also contributing a .travis.yml for continuous integration. Because who does not like continuous integration?

And travis [supports it](https://github.com/travis-ci/travis-ci/issues/1100#issuecomment-160169121) now (tested on another project of mine which uses go-fuse).

chmduquesne:
Ok, sorry. Are you interested in the travis contribution? I can take out
the read modification and make another PR.

hanwen:
can you give me more background on travis? Does this incur any setup and if so, what ? Please open a separate PR for this, BTW.

PR #136, opened 2016-10-19
Change defaultNode.GetAttr() to call file.GetAttr() if not nil

JonathonReinhart:
This enhances nodes which embed `defaultNode` (which is most of them), allowing those implementations to omit a `GetAttr()` method. Without this, when the kernel calls `GetAttr()` for an open file, FUSE will return a size of zero, causing a `read()` of the file to always return 0, even though fuse returns a nonzero number of bytes.

### DIFF
--- a/fuse/nodefs/defaultnode.go
+++ b/fuse/nodefs/defaultnode.go
@@ -126,6 +126,9 @@ func (n *defaultNode) ListXAttr(context *fuse.Context) (attrs []string, code fus
 }
 
 func (n *defaultNode) GetAttr(out *fuse.Attr, file File, context *fuse.Context) (code fuse.Status) {
+	if file != nil {
+		return file.GetAttr(out)
+	}
 	if n.Inode().IsDir() {
 		out.Mode = fuse.S_IFDIR | 0755
 	} else {
--- a/fuse/test/defaultnode_test.go
+++ b/fuse/test/defaultnode_test.go
@@ -1,0 +1,78 @@
+// Copyright 2016 the Go-FUSE Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package test
+
+import (
+	"io/ioutil"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/hanwen/go-fuse/fuse"
+	"github.com/hanwen/go-fuse/fuse/nodefs"
+	"github.com/hanwen/go-fuse/internal/testutil"
+)
+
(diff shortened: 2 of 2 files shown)


JonathonReinhart:
Interesting. The project README says:
```
DISCLAIMER

This is not an official Google product.
```

Anyway, I signed the CLA.

JonathonReinhart:
Yes. I was looking for tests that already exercise `defaultNode`, but didn't find any.

As for testing this, I'm thinking:
- Create a nodefs filesystem that only overrides `Open`, and leaves `GetAttr` to `defaultNode`
- Mount this filesystem, and verify:
   - `stat foo` shows `st_size == 0`   (here `file == nil`)
   - `cat foo` is the expected length/content  (here `file != nil`)

That was the case that I was testing in [`gitlab-fuse`](https://github.com/JonathonReinhart/gitlab-fuse), anyway.

JonathonReinhart:
Hmm, I'm very confused now. My test (https://github.com/hanwen/go-fuse/pull/136/commits/3c402a950453ee44e900fe5c7843238079a648aa) is failing.

When I went back to my other project (for which I'm still in the process of developing automated tests), I saw that even my `GetAttr()` implementation (https://github.com/JonathonReinhart/gitlab-fuse/commit/abb48d25fc75ccf14d25588c17cb062891709474) was no longer working, and I had to revert it (https://github.com/JonathonReinhart/gitlab-fuse/commit/354dd73785d685bc9261dfeaf91f8fdb83b06e1d). This problem came to be when I switched from the `nodefs.MountRoot` wrapper, to manually creating the `FileSystemConnector` to set more options (https://github.com/JonathonReinhart/gitlab-fuse/commit/0af2b0c702f3b1030f36fce6c7fd2d9a03190f88).

So I'm not sure what changes the behavior but it seems to be something to do with FUSE options. :fearful: Still digging.

JonathonReinhart:
It turns out that this patch is only useful if `AttrTimeout` is zero. See `JonathonReinhart/gitlab-fuse#2`. Are you still interested in the patch?

hanwen:
I think the patch is useful. Can you make it work and squash the commits together? Thanks.

JonathonReinhart:
@hanwen Okay, I've updated the test to now pass, and squashed the commits.

hanwen:
merged as https://github.com/hanwen/go-fuse/commit/994fc0fe21ffd9839a391d6513316bf2ef3d0f59

JonathonReinhart:
:+1:

PR #141, opened 2016-11-25
Allow MaxReadAhead to be set - fixes #139

### LINKED ISSUE #139: Can't set max_readahead
I've tried

```go
	mountOpts = &fuse.MountOptions{
		AllowOther:    allowOther,
		Debug:         debugFUSE,
	}
	var opts []string
	opts = append(opts, fmt.Sprintf("max_readahead=%d", maxReadAhead))
	mountOpts.Options = opts
```

But I just get this error

```
/bin/fusermount: mount failed: Invalid argument
```

If I strace the mount I can see it looks correct

```
[pid 32651] execve("/bin/fusermount", ["/bin/fusermount", "/home/ncw/mnt/tmp", "-o", "max_readahead=131072,fsname=local:/tmp,subtype=rclone"], [/* 1 var */] <unfinished ...>
```

Any ideas?

Apologies in advance if this is a stupid question!

ncw:


fuse/opcode.go
```diff
@@ -99,6 +99,9 @@ func doInit(server *Server, req *request) {
 		CongestionThreshold: uint16(server.opts.MaxBackground * 3 / 4),
 		MaxBackground:       uint16(server.opts.MaxBackground),
 	}
+	if server.opts.MaxReadAhead != 0 {
+		out.MaxReadAhead = uint32(server.opts.MaxReadAhead)
```
hanwen: shouldn't this value be capped by the input? Can you find out (maybe look at libfuse source code.)

ncw: the libfuse code in do_init looks (in fuse_lowlevel.c) like this

```c
	if (arg->minor >= 6) {
		if (arg->max_readahead < se->conn.max_readahead)
			se->conn.max_readahead = arg->max_readahead;
```

I read that as saying it sets the  max_readahead to the max_readahead if the current value is lower than the desired max_readahead.

So I think that would translate into go as

```go
	if server.opts.MaxReadAhead != 0 && out.MaxReadAhead < uint32(server.opts.MaxReadAhead) {
		out.MaxReadAhead = uint32(server.opts.MaxReadAhead)
	}
```

Thoughts?

ncw: Yes you are right (duh!) - go code should read

```go
if server.opts.MaxReadAhead != 0 &&  uint32(server.opts.MaxReadAhead) < out.MaxReadAhead {
		out.MaxReadAhead = uint32(server.opts.MaxReadAhead)
}
```

Doing this will make it impossible to set a max read ahead of larger than the default value that fuse provides (128k in my tests).

I had a look at the kernel code and it limits the max readahead to this initial value too.

So if you can't set the readahead to larger than the initial value then I can't see the point of this PR any more as I wanted to increase the max read ahead beyond the default value not decrease it!

Anyway I can make that fix and resubmit if you want.

hanwen: I'm still interested in the fix :)

ncw:
I've updated the PR with the fix as discussed.

rfjakob:
This has been merged on Dec 21, 2016 as https://github.com/hanwen/go-fuse/commit/9ae6c46af39654e47b18f628fd8021e9a51603ca . This pull request should be closed.
```
commit 9ae6c46af39654e47b18f628fd8021e9a51603ca
Author:     Nick Craig-Wood <nick@craig-wood.com>
AuthorDate: Fri Nov 25 14:39:11 2016 +0000
Commit:     Han-Wen Nienhuys <hanwen@google.com>
CommitDate: Wed Dec 21 20:44:18 2016 +0100

    Allow MaxReadAhead to be set - fixes #139
```

PR #153, opened 2017-04-16
Flock implementation with test.

### LINKED ISSUE #134: File Locking
I would like to see support for file locking.
If a process requests or releases a file lock on a fuse mounted file this information may be propagated through the go-fuse library.
A file system implementation may then work with the locking information or not - depending on the needs of the implementation.

Actually I'm using [rfjakob/gocryptfs](https://github.com/rfjakob/gocryptfs) which builds on top of go-fuse.
Once this feature is implemented in go-fuse, gocryptfs could take advatage of it: rfjakob/gocryptfs#39

OneOfOne:
Please let me know if it doesn't work as expected.

Fixes #134

### DIFF
--- a/fuse/test/flock_test.go
+++ b/fuse/test/flock_test.go
@@ -1,0 +1,63 @@
+// Copyright 2016 the Go-FUSE Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build linux
+
+package test
+
+import (
+	"os"
+	"os/exec"
+	"regexp"
+	"syscall"
+	"testing"
+)
+
+func TestFlock(t *testing.T) {
+	tc := NewTestCase(t)
+	defer tc.Cleanup()
+
+	contents := []byte{1, 2, 3}
+	tc.WriteFile(tc.origFile, []byte(contents), 0700)
+
+	f, err := os.OpenFile(tc.mountFile, os.O_WRONLY, 0)
+	if err != nil {
+		t.Fatalf("OpenFile(%q): %v", tc.mountFile, err)
+	}
+	defer f.Close()
+
+	if err = syscall.Flock(int(f.Fd()), syscall.LOCK_EX); err != nil {
+		t.Errorf("Flock returned: %v", err)
+		return
+	}
+
+	if out, err := runExternalFlock(tc.mountFile); err != nil {
+		re := regexp.MustCompile(`flock_test.go:\d+: (.*?)$`) // don't judge me
+		t.Errorf("runExternalFlock(%q): %s", tc.mountFile, re.Find(out))
+	}
+}
+
+func TestFlockExternal(t *testing.T) {
+	fname := os.Getenv("LOCKED_FILE")
+	if fname == "" {
+		t.SkipNow()
+	}
+
+	f, err := os.Open(fname)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer f.Close()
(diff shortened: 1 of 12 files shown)


fuse/test/flock_test.go
```diff
@@ -0,0 +1,63 @@
...
+		t.Fatalf("OpenFile(%q): %v", tc.mountFile, err)
+	}
+	defer f.Close()
+
+	if err = syscall.Flock(int(f.Fd()), syscall.LOCK_EX); err != nil {
+		t.Errorf("Flock returned: %v", err)
+		return
+	}
+
+	if out, err := runExternalFlock(tc.mountFile); err != nil {
```
hanwen: is there no lighter weight mechanism for this? There must be some command-line utility that exposes flock?

OneOfOne: There is the [flock](https://linux.die.net/man/1/flock) command but it's part of util-linux-ng, I don't think it's very portable.

hanwen: I'd rather use that, and skip the test if we can't find the binary. Running the test binary from within the test binary sounds like it might easily break

OneOfOne:
@hanwen did you get a chance to look at it?

hanwen:
squashed and committed as https://github.com/hanwen/go-fuse/commit/ae2e1d5cbf43f14ae1f10885f0b27b9b140480cc

rfjakob:
I think there is still something missing. I have mounted a loopback filesystem with `-debug` and ran flock in the mountpoint: 
```
$ flock --verbose test.lock sleep 1
flock: getting lock took 0.000005 seconds
flock: executing sleep
```
loopback debug output:
```
21:10:32.202610 Dispatch 244: LOOKUP, NodeId: 1. names: [test.lock] 10 bytes
21:10:32.202679 Serialize 244: LOOKUP code: OK value: {NodeId: 4 Generation=2 EntryValid=1.000 AttrValid=1.000 Attr={M0100664 SZ=0 L=1 1026:1026 B0*4096 i0:8782516 A 1493752213.917942455 M 1493752213.917942455 C 1493752213.917661210}}
21:10:32.202791 Dispatch 245: OPEN, NodeId: 4. data: {O_RDONLY,0x8000} 
21:10:32.202834 Serialize 245: OPEN code: OK value: {Fh 3 }
21:10:33.203963 Dispatch 246: FLUSH, NodeId: 4. data: {Fh 3} 
21:10:33.204022 Serialize 246: FLUSH code: OK value: 
21:10:33.204236 Dispatch 247: FLUSH, NodeId: 4. data: {Fh 3} 
21:10:33.204249 Serialize 247: FLUSH code: OK value: 
21:10:33.204308 Dispatch 248: RELEASE, NodeId: 4. data: {Fh 3 0x8000  L0} 
21:10:33.204324 Serialize 248: RELEASE code: OK value: 
```
As you can see, no FLOCK call! The kernel does not seem to pass the call to us. Maybe we have to set a capability bit when mounting?

rfjakob:
Ok I think what is missing is that the flock-related operations (`_OP_GETLK` etc.) are handled in [opcode.go](https://github.com/hanwen/go-fuse/blob/master/fuse/opcode.go)

hanwen:
There was a test (which I didn't run but looked plausible). Are you saying
it's completely broken?

Op di 2 mei 2017 21:25 schreef rfjakob <notifications@github.com>:

rfjakob:
Yes it is missing the kernel interface. But, it does no harm, the new code is just never called at the moment.

The test passes because the kernel keeps track of the locks when the filesystem does not support locking.

Issue #138, opened 2016-11-02
unionfs test race.


hanwen:
Read at 0x00c4204984b0 by goroutine 66:
  github.com/hanwen/go-fuse/unionfs.TestXAttrCaching()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/unionfs/unionfs_xattr_test.go:120 +0x12a0
  testing.tRunner()
      /usr/lib/google-golang/src/testing/testing.go:610 +0xc9

Previous write at 0x00c4204984b0 by goroutine 6:
  github.com/hanwen/go-fuse/unionfs.(*TestFS).GetXAttr()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/unionfs/unionfs_xattr_test.go:35 +0x121
  github.com/hanwen/go-fuse/unionfs.getXAttr()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/unionfs/cachingfs.go:71 +0x110
  github.com/hanwen/go-fuse/unionfs.NewCachingFileSystem.func4()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/unionfs/cachingfs.go:102 +0x62
  github.com/hanwen/go-fuse/unionfs.(*TimedCache).GetFresh()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/unionfs/timedcache.go:76 +0x5f
  github.com/hanwen/go-fuse/unionfs.(*TimedCache).Get()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/unionfs/timedcache.go:55 +0x156
  github.com/hanwen/go-fuse/unionfs.(*cachingFileSystem).GetXAttr()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/unionfs/cachingfs.go:127 +0xe7
  github.com/hanwen/go-fuse/unionfs.(*unionFS).GetXAttr()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/unionfs/unionfs.go:727 +0x13e
  github.com/hanwen/go-fuse/fuse/pathfs.(*pathInode).GetXAttr()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/fuse/pathfs/pathfs.go:393 +0xc3
  github.com/hanwen/go-fuse/fuse/nodefs.(*rawBridge).GetXAttrData()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/fuse/nodefs/fsops.go:402 +0xbb
[...]

PR #157, opened 2017-04-19
fix first unionfs test datarace

jmhodges:
This first race was easy to find and fix in the tests themselves.

It exposes another race in nodefs.FileSystemConnector where a node's
mount (and therefore, treeLock) is is not the same as the grand-parent
node's is.

Updates #138.


hanwen:
thanks!

PR #163, opened 2017-06-29
Do not fork a new process to unmount if user is privileged

pooya:
The root user can issue the umount syscall and does not need the
fusermount binary. Therefore, it is not necessary to fork and exec a new
process just to run the umount binary.

Golang has a global lock held for forking (See exec_unix.go in the golang
source tree):
"Acquire the fork lock so that no other threads create new fds that are
not yet close-on-exec before we fork."

So it would be best if processes would refrain from forking
unnecessarily.

### DIFF
--- a/fuse/mount_linux.go
+++ b/fuse/mount_linux.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"os/exec"
 	"path"
-	"path/filepath"
 	"strings"
 	"syscall"
 	"unsafe"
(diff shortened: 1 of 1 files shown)


pooya:
My employer (Arista) has signed the CLA.
Thanks.

hanwen:
the code looks fine. Is there a way you can add your Arista identity to your GH profile?

pooya:
The email is now linked to my Github profile.

hanwen:
I still can't see it; does it need some button to be clicked?

pooya:
The membership should now be visible on my github profile.
Cheers.

Issue #166, opened 2017-07-14
data race


hanwen:
==================
WARNING: DATA RACE
Write at 0x00c4200d0060 by goroutine 10:
  runtime.mapdelete()
      /home/hanwen/sdk/go1.9beta2/src/runtime/hashmap.go:635 +0x0
  github.com/hanwen/go-fuse/fuse/nodefs.(*Inode).rmChild()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/fuse/nodefs/inode.go:214 +0x17d
  github.com/hanwen/go-fuse/fuse/nodefs.(*FileSystemConnector).forgetUpdate()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/fuse/nodefs/fsconnector.go:157 +0x4fb
  github.com/hanwen/go-fuse/fuse/nodefs.(*rawBridge).Forget()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/fuse/nodefs/fsops.go:128 +0x4c
  github.com/hanwen/go-fuse/fuse.doForget()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/fuse/opcode.go:250 +0x109
  github.com/hanwen/go-fuse/fuse.(*Server).handleRequest()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/fuse/server.go:400 +0x3c5
  github.com/hanwen/go-fuse/fuse.(*Server).loop()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/fuse/server.go:377 +0x1ba

Previous read at 0x00c4200d0060 by goroutine 9:
  runtime.mapiterinit()
      /home/hanwen/sdk/go1.9beta2/src/runtime/hashmap.go:709 +0x0
  github.com/hanwen/go-fuse/fuse/nodefs.(*Inode).Parent()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/fuse/nodefs/inode.go:97 +0xc0
  github.com/hanwen/go-fuse/fuse/pathfs.(*pathInode).GetPath()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/fuse/pathfs/pathfs.go:285 +0x151
  github.com/hanwen/go-fuse/fuse/pathfs.(*pathInode).GetAttr()
      /home/hanwen/go/src/github.com/hanwen/go-fuse/fuse/pathfs/pathfs.go:594 +0x6b
[...]

PR #167, opened 2017-07-14
nodefs: fix data race in Inode.Parent()

rfjakob:
We have to take the treeLock before looking
at the parents map.

However, adding the lock here shows deadlocks when testing with
fsstress (backtrace below [1]).

Dropping pathLock from pathInode.GetPath fixes the deadlocks.
The lock does not seem to  protect anything and fsstress testing runs
fine without it.

Fixes the data race reported at https://github.com/hanwen/go-fuse/issues/166 .

[1] Appendix:
```
goroutine 11 [semacquire]:
[...]
sync.(*Mutex).Lock(0xc4200ee6d8)
	/usr/local/go/src/sync/mutex.go:87 +0x9d
github.com/hanwen/go-fuse/fuse/nodefs.(*FileSystemConnector).forgetUpdate(0xc42001dc80, 0x29, 0x1)
	/home/jakob/go/src/github.com/hanwen/go-fuse/fuse/nodefs/fsconnector.go:137 +0xad
github.com/hanwen/go-fuse/fuse/nodefs.(*rawBridge).Forget(0xc42001dc80, 0x29, 0x1)
	/home/jakob/go/src/github.com/hanwen/go-fuse/fuse/nodefs/fsops.go:128 +0x3f
github.com/hanwen/go-fuse/fuse.doBatchForget(0xc4200921c0, 0xc4200d86c0)
	/home/jakob/go/src/github.com/hanwen/go-fuse/fuse/opcode.go:275 +0x10a
[...]

goroutine 13 [semacquire]:
[...]
sync.(*RWMutex).Lock(0xc420088c08)
	/usr/local/go/src/sync/rwmutex.go:91 +0x6e
github.com/hanwen/go-fuse/fuse/pathfs.(*pathInode).OnRemove(0xc42028f0e0, 0xc42042e460, 0xc42015f680, 0x12)
	/home/jakob/go/src/github.com/hanwen/go-fuse/fuse/pathfs/pathfs.go:341 +0x63
github.com/hanwen/go-fuse/fuse/nodefs.(*Inode).rmChild(0xc42042e460, 0xc42015f680, 0x12, 0xc42004be68)
	/home/jakob/go/src/github.com/hanwen/go-fuse/fuse/nodefs/inode.go:218 +0x18f
github.com/hanwen/go-fuse/fuse/nodefs.(*FileSystemConnector).forgetUpdate(0xc42001dc80, 0x11, 0x1)
[...]

PR #171, opened 2017-07-16
nodefs: fix data race in Inode.Parent()

rfjakob:
Inode.Parent() has to take the treeLock before looking
at the parents map.

Fixes https://github.com/hanwen/go-fuse/issues/166


rfjakob:
(This is https://github.com/hanwen/go-fuse/pull/167 rebased to master)

Issue #175, opened 2017-07-27
getdents: d_ino field not populated


rfjakob:
At the moment, the d_ino field of getdents(2) is always set to `FUSE_UNKNOWN_INO = 4294967295`. This breaks `find -inum`.

In a loopback mount:
```
$ find . -inum 8654086
(returns empty)
$ ls -li
total 0
8654086 -rw-rw-r--. 1 jakob jakob 0 27. Jul 23:23 a
[...]
```
Let's look at what getdents returns:
```
$ strace -e getdents -v ls
getdents(3, [
{d_ino=4294967295, d_off=1, d_reclen=24, d_name="b", d_type=DT_REG},
{d_ino=4294967295, d_off=2, d_reclen=24, d_name="a", d_type=DT_REG},
{d_ino=4294967295, d_off=3, d_reclen=24, d_name="c", d_type=DT_REG},
{d_ino=4294967295, d_off=4, d_reclen=24, d_name=".", d_type=DT_DIR},
{d_ino=4294967295, d_off=5, d_reclen=24, d_name="..", d_type=DT_DIR}], 32768) = 120
getdents(3, [], 32768)                  = 0
a  b  c
```
find uses the bogous d_ino field from getdents() to avoid a stat() call.

hanwen:
tricky: should these be the inode numbers that we return as handls for the kernel (probably: no, because we can never make the lookup semantics work). Then the inodes should be some other number, provided by the user filesystem.

rfjakob:
I'd suggest extending [`struct DirEntry`](https://github.com/hanwen/go-fuse/blob/master/fuse/direntry.go#L20) with an `Ino` field that allows `ClientInodes = true` filesystems to set an inode number in OpenDir.

With `ClientInodes = false`, I would just keep things as-is for now. These filesystem probably care less about POSIX compliance.

rfjakob:
Don't know yet how to do this without breaking all existing users, though.

hanwen:
ClientInodes is a property of the path filesystem. It would be best to put this in nodefs instead.

Issue #96, opened 2016-01-25
unionfs self tests fail 20% of the time on Linux 4.4


rfjakob:
Unionfs seems to be racy. About 20% of the test runs fail with different errors (example output pasted below).
I am on kernel 4.4 now, and given that I have already seen this on older kernels as well ( https://github.com/hanwen/go-fuse/issues/67 ) I suspect a logic error somewhere in unionfs.

```
$ go test github.com/hanwen/go-fuse/unionfs
2016/01/25 23:15:54 Inode = "status/debug_setting" (autoUnionFs(/tmp/419665521/store))
2016/01/25 23:15:54 Inode = "status/debug_setting" (autoUnionFs(/tmp/419665521/store))
2016/01/25 23:15:54 Serialize: SYMLINK code: OK value: 
2016/01/25 23:15:54 Dispatch: READLINK, NodeId: 4. 
2016/01/25 23:15:54 Inode = "status/debug_setting" (autoUnionFs(/tmp/419665521/store))
2016/01/25 23:15:54 Serialize: READLINK code: OK value:  "1"
2016/01/25 23:15:54 Dispatch: BATCH_FORGET, NodeId: 0. data: {2}  32 bytes
2016/01/25 23:15:54 Dispatch: FORGET, NodeId: 1. data: {1} 
2016/01/25 23:15:54 Found version: version
2016/01/25 23:15:54 Adding workspace manual1 for roots UnionFs([LoopbackFs(/tmp/851926190/store/backing1) cachingFileSystem(LoopbackFs(/tmp/851926190/store/backing1/READONLY))])
2016/01/25 23:15:54 Looking for new filesystems
2016/01/25 23:15:54 Adding workspace backing1 for roots UnionFs([LoopbackFs(/tmp/851926190/store/backing1) cachingFileSystem(LoopbackFs(/tmp/851926190/store/backing1/READONLY))])
2016/01/25 23:15:54 Done looking
2016/01/25 23:15:54 Looking for new filesystems
2016/01/25 23:15:54 Adding workspace backing1 for roots UnionFs([LoopbackFs(/tmp/657096757/store/backing1) cachingFileSystem(LoopbackFs(/tmp/657096757/store/backing1/READONLY))])
2016/01/25 23:15:54 Done looking
2016/01/25 23:15:54 Looking for new filesystems
[...]

PR #177, opened 2017-07-29
Add support for d_ino

rfjakob:
Fixes https://github.com/hanwen/go-fuse/issues/175 while hopefully causing little breakage for apps that use go-fuse. As long as they embed `NewDefaultNode` (nodefs) / `NewDefaultFileSystem` (pathfs) apps should see no change in behavoir.

The patch series passes the go-fuse test suite ([most of the time](https://github.com/hanwen/go-fuse/issues/96), anyway) and the unmodified gocryptfs test suite.

### DIFF
--- a/fuse/direntry.go
+++ b/fuse/direntry.go
@@ -30,6 +30,24 @@ func (d DirEntry) String() string {
 	return fmt.Sprintf("%o: %q", d.Mode, d.Name)
 }
 
+// DirEntryIno is a DirEntry struct extended by the "Ino" field for the
+// inode number.
+type DirEntryIno struct {
+	// Mode is the file's mode. Only the high bits (eg. S_IFDIR)
+	// are considered.
+	Mode uint32
+
+	// Name is the basename of the file in the directory.
+	Name string
+
+	// Ino is the inode number
+	Ino uint64
+}
+
+func (d DirEntryIno) String() string {
+	return fmt.Sprintf("%o: %q, ino=%d", d.Mode, d.Name, d.Ino)
+}
+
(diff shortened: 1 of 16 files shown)


rfjakob:
Limitation: The `..` and `.` directory entries still get `FUSE_UNKNOWN_INO` (set [here](https://github.com/hanwen/go-fuse/blob/55260d4a541176d11a44cfab2bcd9cc962bd71ab/fuse/nodefs/fsops.go#L157)). This looks harder the fix and less important.

fuse/direntry.go
```diff
@@ -30,6 +30,24 @@ func (d DirEntry) String() string {
 	return fmt.Sprintf("%o: %q", d.Mode, d.Name)
 }
 
+// DirEntryIno is a DirEntry struct extended by the "Ino" field for the
+// inode number.
+type DirEntryIno struct {
```
hanwen: why? can't you simply add the Ino field to DirEntry? We could translate the default value 0 to FUSE_UNKNOWN_INODE.

rfjakob: Yes, that would make things much easier. But DirEntry is public API, and changing it may break users.

rfjakob: I mean, I'm good with that, but I would not propose it without consulting with you first!

hanwen: It only breaks people that dont use field names. They can fix up themselves.

rfjakob:
Ok, good. Will rework!

rfjakob:
Superseded by https://github.com/hanwen/go-fuse/pull/180 .

PR #180, opened 2017-08-05
fuse, loopback: return actual inode numbers from READDIR

rfjakob:
Supersedes https://github.com/hanwen/go-fuse/pull/177. Passes both go-fuse's `all.bash` and gocryptfs' `test.bash` (unmodified). Single commit message follows:

--------------------------

When an app in a FUSE mount calls getdents(2), go-fuse receives
READDIR[PLUS] and calls the filesystem's OpenDir function that
returns []DirEntry.

The data returned from getdents(2) contains an inode number for
each directory entry, "d_ino". Until now, struct DirEntry had no
corresponding field and the value passed to the kernel was always
FUSE_UNKNOWN_INO = 0xffffffff

This broke apps that actually look at the d_ino field, like
"find -inum".

This commit adds the "Ino" filed to struct DirEntry. If the field
is not set by the filesystem, it is set to FUSE_UNKNOWN_INO,
as before. Otherwise it is left alone and passed to the kernel.

loopbackFileSystem's OpenDir function is extended to set the inode
number. A test verifies that the returned inode number is sane.

Fixes https://github.com/hanwen/go-fuse/issues/175

### DIFF
--- a/fuse/test/loopback_linux_test.go
+++ b/fuse/test/loopback_linux_test.go
@@ -5,11 +5,14 @@
 package test
 
 import (
+	"encoding/binary"
 	"io/ioutil"
 	"os"
 	"syscall"
 	"testing"
 	"time"
+
+	"github.com/hanwen/go-fuse/fuse"
 )
 
 func TestTouch(t *testing.T) {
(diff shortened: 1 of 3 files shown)


fuse/test/loopback_linux_test.go
```diff
@@ -157,3 +160,34 @@ func TestSpecialEntries(t *testing.T) {
...
+	}
+	defer d.Close()
+	buf := make([]byte, 100)
+	// readdir(3) use getdents64(2) internally which returns linux_dirent64
+	// structures. We don't have readdir(3) so we call getdents64(2) directly.
+	n, err := syscall.Getdents(int(d.Fd()), buf)
+	if n == 0 {
+		t.Error("empty directory - we need at least one file")
+	}
+	// The inode number of the first directory entry (=hello.txt)
```
hanwen: is hello.txt guaranteed to be the first? I think it may be an artifact of the ordering of go-fuse itself. CAn you collect all the entries and compare the one with hello.txt?

rfjakob: Actually, it does not matter *which* file we get. The test creates `hello.txt` so there is *at least* one file, regardless of which tests were executed before.

rfjakob: Oh, disregard that. Will fix!

rfjakob:
Fixed in https://github.com/hanwen/go-fuse/pull/181

PR #181, opened 2017-08-08
fuse, loopback: return actual inode numbers from READDIR

rfjakob:
When an app in a FUSE mount calls getdents(2), go-fuse receives
READDIR[PLUS] and calls the filesystem's OpenDir function that
returns []DirEntry.

The data returned from getdents(2) contains an inode number for
each directory entry, "d_ino". Until now, struct DirEntry had no
corresponding field and the value passed to the kernel was always
FUSE_UNKNOWN_INO = 0xffffffff

This broke apps that actually look at the d_ino field, like
"find -inum".

This commit adds the "Ino" filed to struct DirEntry. If the field
is not set by the filesystem, it is set to FUSE_UNKNOWN_INO,
as before. Otherwise it is left alone and passed to the kernel.

loopbackFileSystem's OpenDir function is extended to set the inode
number. A test verifies that the returned inode number is sane.

Fixes https://github.com/hanwen/go-fuse/issues/175

### DIFF
--- a/fuse/test/loopback_linux_test.go
+++ b/fuse/test/loopback_linux_test.go
@@ -10,6 +10,8 @@ import (
 	"syscall"
 	"testing"
 	"time"
+
+	"github.com/hanwen/go-fuse/fuse"
 )
 
 func TestTouch(t *testing.T) {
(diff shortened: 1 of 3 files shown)


fuse/test/loopback_linux_test.go
```diff
@@ -157,3 +159,44 @@ func TestSpecialEntries(t *testing.T) {
...
+	}
+	buf = buf[:n]
+	entries := parseDirents(buf)
+	t.Logf("parseDirents returned %d entries", len(entries))
+	// Find "hello.txt" and check inode number.
+	for _, entry := range entries {
+		if entry.name != filename {
+			continue
+		}
+		if entry.ino != 0 && entry.ino != fuse.FUSE_UNKNOWN_INO {
```
hanwen: this should be the inode number of the underlying loopback file, no?

regardless, I'll merge tihs now.

rfjakob: That would be a stronger check, yes. Will follow up.

PR #184, opened 2017-09-08
Initialize the loops waitGroup when initializing the server

pooya:
Otherwise, the Serve() and Unmount() calls will race calling
loops.Add(1) and loops.Wait() and the outcome is undefined.

### DIFF
--- a/fuse/server.go
+++ b/fuse/server.go
@@ -187,6 +187,10 @@ func NewServer(fs RawFileSystem, mountPoint string, opts *MountOptions) (*Server
 		// TODO - unmount as well?
 		return nil, fmt.Errorf("init: %s", code)
 	}
+	// Make sure Unmount() will wait for a corresponding Serve() to finish;
+	// Initializing the waitGroup cannot happen in Serve() as the caller can
+	// never be sure loop is initialized before calling Unmount()
+	ms.loops.Add(1)
 	return ms, nil
 }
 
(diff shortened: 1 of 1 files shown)


hanwen:
your two pull requests seem conflicting: if you are always calling WaitMount, then you would always call Unmount after WaitMount. Since WaitMount waits for Serve to process the first request, there is no race.

pooya:
The main purpose of this patch is to have a way to correctly schedule a umount to be called after a hard deadline; to unmount and kill any run-away servers.
If there is something wrong with the fuse server, the WaitMount() itself might hang. So I don't think it is the best idea to start such a timer after WaitMount() returns.

hanwen:
can you specify more exactly what "something wrong" means? If your server is broken and in an unknown state, then is there any sense in worrying about deadlines?

pooya:
This is not an issue for the production use-case where the server's state is unknown. We are trying to run a server in a test and we have a timeout to kill the server (and fail the test) if it runs for too long. Using golang's -race option correctly reports this race between calling Serve() and Unmount().

hanwen:
I think the real problem is that the API is designed in the wrong way. 

The Serve loop should be started from NewServer. Possibly, it should even do WaitMount implicitly. 

Then there should be a Server.Wait() which waits until the serve loop exits.

pooya:
Yes. But changing the API in that way would be backwards-incompatible I assume.
Of course, up to you to decide whether to incorporate this change or not, not a big issue.

hanwen:
Let me see what kind of problems changing the API would create.

PR #193, opened 2017-10-31
Consult the input.Size in order to differentiate two different opcode…

pooya:
…s of [get|list]attr

Consulting the output size is incorrect as it does not take into account the
cases where an xattr with an empty value exists.

Moreover, treat listxattr exactly the same as getxattr as the two both
can be used for two different operations.

### DIFF
--- a/fuse/request.go
+++ b/fuse/request.go
@@ -198,11 +198,13 @@ func (r *request) serializeHeader(flatDataSize int) (header []byte) {
 		dataLength = 0
 	}
 
-	// GETXATTR is two opcodes in one: get xattr size (return
-	// structured GetXAttrOut, no flat data) and get xattr data
+	// [GET|LIST]XATTR is two opcodes in one: get/list xattr size (return
+	// structured GetXAttrOut, no flat data) and get/list xattr data
 	// (return no structured data, but only flat data)
-	if r.inHeader.Opcode == _OP_GETXATTR && flatDataSize > 0 {
-		dataLength = 0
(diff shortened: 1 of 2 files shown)


fuse/request.go
```diff
@@ -198,11 +198,13 @@ func (r *request) serializeHeader(flatDataSize int) (header []byte) {
...
 	}
 
-	// GETXATTR is two opcodes in one: get xattr size (return
-	// structured GetXAttrOut, no flat data) and get xattr data
+	// [GET|LIST]XATTR is two opcodes in one: get/list xattr size (return
+	// structured GetXAttrOut, no flat data) and get/list xattr data
 	// (return no structured data, but only flat data)
-	if r.inHeader.Opcode == _OP_GETXATTR && flatDataSize > 0 {
-		dataLength = 0
+	if r.inHeader.Opcode == _OP_GETXATTR || r.inHeader.Opcode == _OP_LISTXATTR {
```
hanwen: this looks credible, but it would be great if you could add a test so this doesn't regress in the future.

pooya: The commit is fixing two bugs:
1. Not respecting the two types of listxattrs
2. consulting the output size instead of the input size

The second issue is easy to miss, so a test was added for it.

I don't see a test for verifying the second opcode of GetXAttr is working correctly and I assume writing such a test in golang would be more tricky as the syscalls do not accept a size.

hanwen: re: syscalls: yeah, that is actually my fault. I added the XAttr syscalls API to the syscall package and tried to make it easy to use, which was probably a mistake.

hanwen:
thanks, I've squashed and pushed this.

PR #194, opened 2017-11-07
Limit the length of the file names to 255 characters

pooya:
Just like almost all other filesystems.

### DIFF
--- a/fuse/request.go
+++ b/fuse/request.go
@@ -13,6 +13,8 @@ import (
 	"unsafe"
 )
 
+const MAX_FILENAME = 255
+
 var sizeOfOutHeader = unsafe.Sizeof(OutHeader{})
 var zeroOutBuf [outputHeaderSize]byte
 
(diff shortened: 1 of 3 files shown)


rfjakob:
I don't think this is a good idea. Ntfs has a 260-character limit for example. This change means you can't write an ntfs driver with go-fuse.

rfjakob:
Actually, the NTFS limit seems to be 255 **unicode characters** ( https://msdn.microsoft.com/en-us/library/ee681827(VS.85).aspx#limits ). Still, limiting to 255 **bytes** would break NTFS.

Cryptomator and CryFS, which use libfuse, also support more than 255 bytes ( https://nuetzlich.net/gocryptfs/comparison/#file-names )

pooya:
We have to option of switching to RuneCountInString [1] if we want to support unicode.
In general, I agree there may exist a filesystem that wants to support more than 255, but it looks like CryFS is in C++ and cryptomator is in Java according to the link you provided.

[1] https://golang.org/pkg/unicode/utf8/#RuneCountInString

hanwen:
what problem is this solving? Either linux VFS enforces the limit, in which case FUSE never gets the request, or it allows it, and then it is up to the FS author to decide if what length they want to support.

pooya:
I can modify the patch to accept a MAX_LENGTH config option instead of using 255. The reason it was not introduce as part of the first iteration was that I did not think anyone would be interested in any other limit.

hanwen:
I am rejecting this. Please open a bug report if therre is an underlying problem.

PR #199, opened 2018-01-16
implement file locking: getlk/setlk/setlkw/flock

### LINKED ISSUE #170: flock implementation is untested
see https://github.com/hanwen/go-fuse/pull/153 

The kernel ops are not hooked up, so the test just exercises the kernel's default locking.

On a related note, on my ubuntu 14.04 release, I get

--- FAIL: TestFlock (0.02s)
	flock_test.go:40: runExternalFlock("/tmp/TestFlock737947072/mnt/hello.txt"): /usr/bin/flock: unrecognized option '--verbose'
		
		Usage:
		 /usr/bin/flock [-sxun][-w #] fd#
		 /usr/bin/flock [-sxon][-w #] file [-c] command...
		 /usr/bin/flock [-sxon][-w #] directory [-c] command...
		
		Options:
		 -s  --shared     Get a shared lock
		 -x  --exclusive  Get an exclusive lock
		 -u  --unlock     Remove a lock
		 -n  --nonblock   Fail rather than wait
		 -w  --timeout    Wait for a limited amount of time
		 -o  --close      Close file descriptor before running command
		 -c  --command    Run a single command string through the shell
		 -h  --help       Display this text
		 -V  --version    Display version
		
		 (<nil>)
FAIL

kaoet:
Fix #170


kaoet:
I signed it!

fuse/nodefs/api.go
```diff
@@ -102,6 +102,11 @@ type Node interface {
 	SetXAttr(attr string, data []byte, flags int, context *fuse.Context) fuse.Status
 	ListXAttr(context *fuse.Context) (attrs []string, code fuse.Status)
 
+	// File locking
+	GetLk(file File, owner uint64, lk *fuse.FileLock, flags uint32, padding uint32, out *fuse.FileLock, context *fuse.Context) (code fuse.Status)
```
hanwen: padding should never be part of function signatures.

fuse/test/file_lock_test.go
```diff
@@ -14,8 +14,7 @@ import (
 	"testing"
 )
 
-// See https://github.com/hanwen/go-fuse/issues/170
-func disabledTestFlock(t *testing.T) {
+func TestFlock(t *testing.T) {
```
hanwen: this test was a nice try, but I don't trust it. Can you make something that implements a lock, and verifies that the user-space locking code gets called?

fuse/nodefs/api.go
kaoet: Fixed

fuse/test/file_lock_test.go
kaoet: Fixed

mariash:
Hi, we also tested this with nfs v4 and locking works, so we would appreciate if this PR would be merged.

hanwen:
Hi! Thanks for reminding me. I'll have a look this week

Op 16 mrt. 2018 20:22 schreef "Maria Shaldibina" <notifications@github.com>:

fuse/nodefs/files.go
```diff
@@ -167,12 +167,56 @@ func (f *loopbackFile) Fsync(flags int) (code fuse.Status) {
 	return r
 }
 
-func (f *loopbackFile) Flock(flags int) fuse.Status {
-	f.lock.Lock()
-	r := fuse.ToStatus(syscall.Flock(int(f.File.Fd()), flags))
-	f.lock.Unlock()
+const (
+	F_OFD_GETLK  = 36
```
hanwen: this looks like a syscall interface. Is this the same on OSX ? I would expect this defs to be a linux-specific file (I assume you wrote and tested this under linux.)

fuse/nodefs/api.go
```diff
@@ -102,6 +102,11 @@ type Node interface {
 	SetXAttr(attr string, data []byte, flags int, context *fuse.Context) fuse.Status
 	ListXAttr(context *fuse.Context) (attrs []string, code fuse.Status)
 
+	// File locking
```
hanwen: I know this wasn't documented well before, but could you add some more wording here? I guess Lkw locks for write and Lk is a concurrent (?) read lock? If you don't want to write docs, add a manpage reference.

fuse/nodefs/files.go
```diff
@@ -167,12 +167,56 @@ func (f *loopbackFile) Fsync(flags int) (code fuse.Status) {
...
+		case syscall.F_RDLCK:
+			op = syscall.LOCK_SH
+		case syscall.F_WRLCK:
+			op = syscall.LOCK_EX
+		case syscall.F_UNLCK:
+			op = syscall.LOCK_UN
+		default:
+			return fuse.EINVAL
+		}
+		if !blocking {
```
hanwen: if this is controlled from a field in the flags, why have the separate argument 'blocking' ?

fuse/test/file_lock_test.go
```diff
@@ -0,0 +1,125 @@
+// Copyright 2016 the Go-FUSE Authors. All rights reserved.
```
hanwen: 2018

did you add your name to the AUTHORS file?

fuse/test/file_lock_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	defer f.Close()
+
+	if err = syscall.Flock(int(f.Fd()), syscall.LOCK_EX); err != nil {
+		t.Errorf("Flock returned: %v", err)
+		return
+	}
+
+	if out, err := runExternalFlock(cmd, tc.mountFile); !bytes.Contains(out, []byte("failed to get lock")) {
+		t.Errorf("runExternalFlock(%q): %s (%v)", tc.mountFile, out, err)
```
hanwen: use %q iso. %s. For []byte %s will use [1 2 3] as formatting.

fuse/test/file_lock_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+func runExternalFlock(flockPath, fname string) ([]byte, error) {
+	f, err := os.OpenFile(fname, os.O_WRONLY, 0)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	cmd := exec.Command(flockPath, "--verbose", "--exclusive", "--nonblock", "3")
+	cmd.Env = append(cmd.Env, "LC_ALL=C") // in case the user's shell language is different
+	cmd.ExtraFiles = []*os.File{f}
```
hanwen: this needs some more comment. Why do you not pass the filename to the flock command line?

fuse/test/file_lock_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	go s.Serve()
+	if err := s.WaitMount(); err != nil {
+		t.Fatal("WaitMount", err)
+	}
+	defer s.Unmount()
+
+	node := &lockingNode{
+		Node:    nodefs.NewDefaultNode(),
+		flockInvoked: false,
```
hanwen: drop (is implicit)

fuse/test/file_lock_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	root.Inode().NewChild("foo", false, node)
+
+	realPath := filepath.Join(dir, "foo")
+	cmd:=exec.Command(flock, "--nonblock", realPath, "echo", "locked")
```
hanwen: did you run gofmt?

fuse/test/file_lock_test.go
```diff
@@ -0,0 +1,125 @@
...
+	out, err :=cmd.CombinedOutput()
+	if err!=nil {
+		t.Fatalf("flock %v: %v",err, string(out))
+	}
+	if !node.flockInvoked {
+		t.Fatalf("flock is not invoked")
+	}
```
hanwen: this test makes sure the locking really goes through FUSE (which is good), but I can't see the functionality of GetLk/SetLk/SetLkw tested separately. From the signatures, it seems one can have different locks on ranges of the file, and it would be good to exercise that too. Can you think of a way to add coverage for those features? Maybe you could record the ranges of the locks and verify they are as you expect?

fuse/nodefs/files.go
paulcwarren: We tested these constant on OSX and they seem to be working.

fuse/test/file_lock_test.go
paulcwarren: Not sure who he is but we added his name to the authors file on his behalf.  Hope that's ok with you Kaoet?

fuse/test/file_lock_test.go
paulcwarren: Can you clarify.  By drop (is implicit) do you mean that flockinvoked doesn't need to be set to false because that's the default for a bool or ???

kaoet:
Let's use @paulcwarren 's PR. #220

fuse/nodefs/files.go
hanwen: blocking is determined by SetLk vs SetLkw

fuse/test/file_lock_test.go
hanwen: I missed update to AUTHORs in your commit?

fuse/test/file_lock_test.go
hanwen: ping

fuse/test/file_lock_test.go
hanwen: ping

fuse/test/file_lock_test.go
hanwen: correct. The default value is false, so no need to mention it.

fuse/test/file_lock_test.go
mariash: done

fuse/test/file_lock_test.go
mariash: This test was slightly changed. Since on Ubuntu Trusty flock does not have a verbose option we can't assert an output. So we relaxed an assertion that command fails which in turn removed this `Errorf` line. 

https://github.com/hanwen/go-fuse/pull/220/files#diff-49306d0ec901b97786aec218f18ee349R21

fuse/test/file_lock_test.go
mariash: added in our PR https://github.com/hanwen/go-fuse/pull/220/files#diff-49306d0ec901b97786aec218f18ee349R56

fuse/test/file_lock_test.go
mariash: done in our PR https://github.com/hanwen/go-fuse/pull/220/files#diff-49306d0ec901b97786aec218f18ee349R117

fuse/test/file_lock_test.go
mariash: done

fuse/test/file_lock_test.go
mariash: We added additional testing to ensure SetLk and SetLkw are called but we couldn't find a way to use flock to invoke GetLk. We tried cgo in an attempt to call fcntl.c but cgo isnt allowed in test files. Maybe you have some other ideas?

Issue #204, opened 2018-03-07
MacOS breakage: "osxfuse: vnode changed generation"


rfjakob:
I recenty got a ticket filed against gocryptfs with macos users seeing errors when copying files: https://github.com/rfjakob/gocryptfs/issues/213

I have now managed to get my hands on macos box (high sierra, osxfuse 3.7.1). The [extractloop.bash](https://github.com/rfjakob/gocryptfs/blob/master/tests/stress_tests/extractloop.bash) stress test reproduces the issue, both for gocryptfs (just run `./extractloop.bash`) and loopback (running `./extractloop.bash -loopback`). 

The tar extract fails with lots of
```
Can't create file
```
errors while we get these [osxfuse errors](https://github.com/osxfuse/kext/blob/534773bd3135f04b7f6fb801faa2f8991daf4a6b/osxfuse/fuse_node.c#L225) in dmesg:
```
osxfuse: vnode changed generation
osxfuse: vnode changed generation
osxfuse: vnode changed generation
```
I was tempted to suggest that osxfuse is just completely broken, but `./extractloop.bash -encfs` (running the same tarball extract test against encfs) works fine. So I guess we should be doing something differently on macos. Ideas?

hanwen:
the generation numbes are generated in a harebrained way, 

https://github.com/hanwen/go-fuse/blob/da8b3e44dd6cb8bbb2325ca4135f70cc6398773c/fuse/nodefs/handle.go#L85

rfjakob:
Let's see if the osxfuse people can shed some light on this: https://github.com/osxfuse/osxfuse/issues/482

magnusja:
Looks very much the same for encfs? https://github.com/vgough/encfs/blob/1c273ecba57d95c529f21d72c7615e2bb8a911cf/encfs/Context.cpp#L190

PR #205, opened 2018-03-13
darwin: use simpleHandleMap that never reuses handle numbers

rfjakob:
Implement simpleHandleMap that never reuses handle numbers and
uses a simple map to track handle numbers.

This seems to get rid of the

	osxfuse: vnode changed generation

errors we are seeing on darwin, while sacrificing
some performance.

I have tested simpleHandleMap on darwin and linux,
seems to work fine.

https://github.com/rfjakob/gocryptfs/issues/213
https://github.com/hanwen/go-fuse/issues/204
https://github.com/osxfuse/osxfuse/issues/482

### DIFF
--- a/fuse/nodefs/handle_test.go
+++ b/fuse/nodefs/handle_test.go
@@ -5,18 +5,24 @@
 package nodefs
 
 import (
-	"strings"
+	"os"
 	"testing"
 )
 
-func markSeen(t *testing.T, substr string) {
-	if r := recover(); r != nil {
-		s := r.(string)
-		if strings.Contains(s, substr) {
-			t.Log("expected recovery from: ", r)
-		} else {
-			panic(s)
-		}
+var skipTestHandleMapGeneration = false
+
+func TestMain(m *testing.M) {
+	// Test both handleMap implementations
+	newHandleMap = newPortableHandleMap
+	r := m.Run()
+	if r != 0 {
+		os.Exit(r)
+	}
+	newHandleMap = newSimpleHandleMap
+	skipTestHandleMapGeneration = true
+	r = m.Run()
+	if r != 0 {
+		os.Exit(r)
 	}
 }
 
@@ -24,7 +30,7 @@ func TestHandleMapLookupCount(t *testing.T) {
 	for _, portable := range []bool{true, false} {
 		t.Log("portable:", portable)
 		v := new(handled)
-		hm := newPortableHandleMap()
+		hm := newHandleMap()
 		h1, g1 := hm.Register(v)
 		h2, g2 := hm.Register(v)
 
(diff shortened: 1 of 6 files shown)


fuse/nodefs/handle_test.go
```diff
@@ -5,26 +5,32 @@
...
 )
 
-func markSeen(t *testing.T, substr string) {
-	if r := recover(); r != nil {
-		s := r.(string)
-		if strings.Contains(s, substr) {
-			t.Log("expected recovery from: ", r)
-		} else {
-			panic(s)
-		}
```
rfjakob: The markSeen function was not used anywhere

hanwen:
I have one concern with this, which is that the other FUSE library (rsc's library and bazil.org which was derived from it.) uses the same approach here. Can you confirm that they too see the vnode changed messages?

rfjakob:
I'll try to find out.

Even with simpleHandleMap I see (different) osxfuse messages (something about disappearing nodes), but I don't see how we can code this even more defensively to keep osxfuse happy

rfjakob:
I could not reproduce the problem the bazil-based [keybase/loopback](https://github.com/keybase/loopback). Still getting lots of

    kernel: (osxfuse) osxfuse: disappearing vnode <private> (name=asm type=2 action=80000002)

but no "vnode changed generation" and no errors returned to user-space.

Also, I have a [user report](https://github.com/rfjakob/gocryptfs/issues/213#issuecomment-373181552) that simpleHandleMap makes the problem appear less often but does not fix it, so this seems to be a dead end.

rfjakob:
Plot twist: I had the user re-test the change, and it does [fix the problem for him](https://github.com/rfjakob/gocryptfs/issues/213#issuecomment-373599379):

> Ouch, my bad! I was still using the executable installed by homebrew. I copied 20GB and used the mount all day for all sorts of stuff and I do not have any issues anymore!

This matches what I see on the Mac box I managed to get my hands on.

While it is curious why it does not happen for bazil, MacOS kernel module development seems to be a world of pain - I gave up when I had to edit an sqlite file in recovery mode to get my self-compiled osxfuse.kext to load.

I suggest to merge this workaround that fixes the problem with no obvious downsides.

hanwen:
see https://github.com/hanwen/go-fuse/blob/da8b3e44dd6cb8bbb2325ca4135f70cc6398773c/fuse/nodefs/handle.go#L85 here

we get the generation number from the handleMap, and we increase it when we hand out a new node, but the effect is persisted, ie. in the sequence

   lookup -> new node A, gen  = 1
   lookup -> new node B, gen = 2
   lookup -> existing node A, gen = 2

A will change generation numbers.

Rather than papering the problem over without us understanding what is going on, can we fix the real bug here?

rfjakob:
Oh, now I see it. I'll prepare a patch.

rfjakob:
And bazil does not have this bug because it returns the saved generation number:

https://github.com/bazil/fuse/blob/371fbbdaa8987b715bdd21d6adc4c9b20155f748/fs/serve.go#L489
```go
	if id, ok := c.nodeRef[node]; ok {
		sn := c.node[id]
		sn.refs++
		return id, sn.generation
	}
```

PR #206, opened 2018-03-16
handleMap: track per-handle generation number

rfjakob:
Fixes https://github.com/hanwen/go-fuse/issues/204 properly.


hanwen:
I added a test and pushed.

rfjakob:
Good point, test was missing. Thanks!

PR #208, opened 2018-03-21
nodefs: Make Create call File.GetAttr if possible - fixes #207

### LINKED ISSUE #207: PathFileSystem Create returning EIO to userspace
I've been trying to create a PathFileSystem fuse FS for rclone using go-fuse.

However I can't get [the Create callback](https://github.com/ncw/rclone/blob/go-fuse/cmd/mount2/fs.go#L228-L251) to work.

When I do this

```
$ echo hellohellohello > /mnt/tmp/hello3
bash: /mnt/tmp/hello3: Input/output error
```

I see this

```
2018/03/21 14:45:11 Dispatch 55: LOOKUP, NodeId: 1. names: [hello3] 7 bytes
2018/03/21 14:45:11 Serialize 55: LOOKUP code: 2=no such file or directory value: {NodeId: 0 Generation=0 EntryValid=0.000 AttrValid=0.000 Attr={M00 SZ=0 L=0 0:0 B0*0 i0:0 A 0.000000000 M 0.000000000 C 0.000000000}}
2018/03/21 14:45:11 Dispatch 56: CREATE, NodeId: 1. data: {0100664 [CREAT,WRONLY,TRUNC,0x8000] (02)} names: [hello3] 7 bytes
2018/03/21 14:45:11 DEBUG : hello3: Create: flags=0101101, mode=0100664
2018/03/21 14:45:11 DEBUG : hello3: Open: flags=O_WRONLY|O_CREATE|O_TRUNC|0x8000
2018/03/21 14:45:11 DEBUG : hello3: >Open: fd=hello3 (w), err=<nil>
2018/03/21 14:45:11 DEBUG : hello3: >Create: handle=File fh=0xc42047e550(hello3) () TRUNC,CREAT,WRONLY,0x8000 NONSEEK, status=OK
2018/03/21 14:45:11 status = OK
2018/03/21 14:45:11 Serialize 56: CREATE code: OK value: {NodeId: 11 Generation=10 {M00 SZ=0 L=1 502:502 B0*0 i0:11 A 0.000000000 M 0.000000000 C 0.000000000} &{3 4 0}}
```

Note the OK status in the serialize return (and there were no logs after that).

I conjecture that [this bit of code in the linux kernel source fs/fuse/dir.c](https://github.com/torvalds/linux/blob/master/fs/fuse/dir.c#L452) is noting that the mode of the file is 0 and changing that OK status into EIO.

```c
	err = fuse_simple_request(fc, &args);
	if (err)
		goto out_free_ff;

	err = -EIO;
	if (!S_ISREG(outentry.attr.mode) || invalid_nodeid(outentry.nodeid))
		goto out_free_ff;
```

Which is telling me that I should be setting attributes with the Create call.
I don't see a way of setting that attribute returned with the Create call.  I have implemented `GetAttr` on the handle but it doesn't seem to be being called.

What I think is happening is that
  * [nodefs Create gets called](https://github.com/hanwen/go-fuse/blob/master/fuse/nodefs/fsops.go#L364-L377)
  * this calls [childLookup](https://github.com/hanwen/go-fuse/blob/master/fuse/nodefs/fsconnector.go#L93-L106)
  * this calls [pathfs GetAttr](https://github.com/hanwen/go-fuse/blob/master/fuse/pathfs/pathfs.go#L578)
  * However inserting a few debug statements I find that it never finds an open file
  * so it ends up calling the [fallback GetAttr](https://github.com/hanwen/go-fuse/blob/master/fuse/pathfs/pathfs.go#L596)
  * which returns ENOENT as as far as rclone is concerned the file hasn't been created yet
  * and back in [childLookup the error return is ignored](https://github.com/hanwen/go-fuse/blob/master/fuse/nodefs/fsconnector.go#L95) - I think that is a bug!

So I think this is caused by go-fuse not finding the newly opened file and using that for its GetAttr which I think it should be doing. I think what should happen is that `childLookup` or a variant should call `GetAttr` on the open file handle directly.

I'll make a PR with a proposed fix in a moment

ncw:
Before this fix, Create was calling GetAttr (indirectly) which ends up
doing a GetAttr on the file system, not the file handle.  This is
inefficient and causes problems for clients which haven't added the
open file to the directory by the time the call returns.

After this fix Create via childLookup calls GetAttr on the file handle
directly if available.

### DIFF
--- a/fuse/nodefs/fsconnector.go
+++ b/fuse/nodefs/fsconnector.go
@@ -91,8 +91,18 @@ func (c *FileSystemConnector) verify() {
 }
 
 // childLookup fills entry information for a newly created child inode
-func (c *rawBridge) childLookup(out *fuse.EntryOut, n *Inode, context *fuse.Context) {
-	n.Node().GetAttr((*fuse.Attr)(&out.Attr), nil, context)
+func (c *rawBridge) childLookup(out *fuse.EntryOut, n *Inode, f File, context *fuse.Context) {
+	// use File if present
+	if f != nil {
+		code := f.GetAttr((*fuse.Attr)(&out.Attr))
+		if !code.Ok() {
+			f = nil // fallback to Inode based GetAttr
+		}
+
+	}
+	if f == nil {
+		_ = n.Node().GetAttr((*fuse.Attr)(&out.Attr), nil, context)
+	}
 	n.mount.fillEntry(out)
 	out.NodeId, out.Generation = c.fsConn().lookupUpdate(n)
 	if out.Ino == 0 {
--- a/fuse/nodefs/fsops.go
+++ b/fuse/nodefs/fsops.go
@@ -282,7 +282,7 @@ func (c *rawBridge) Mknod(input *fuse.MknodIn, name string, out *fuse.EntryOut) 
 
 	child, code := parent.fsInode.Mknod(name, input.Mode, uint32(input.Rdev), &input.Context)
 	if code.Ok() {
-		c.childLookup(out, child, &input.Context)
+		c.childLookup(out, child, nil, &input.Context)
 		code = child.fsInode.GetAttr((*fuse.Attr)(&out.Attr), nil, &input.Context)
 	}
 	return code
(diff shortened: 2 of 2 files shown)


hanwen:
in general, you should always have a test that fails before your fix and succeeds with your fix.

ncw:
> in general, you should always have a test that fails before your fix and succeeds with your fix.

OK I'll have a look at the tests and see what I can do!

PR #213, opened 2018-05-01
Changed FromStatfsT on Darwin to fix df output 

bolshevik:
`osxfuse/fuse` provides a possibility to implement a standard and an extended `statfs` calls, the previous mapping is relevant for the `statfs_x` call, which is not implemented in `go-fuse` yet. 

https://github.com/osxfuse/fuse/blob/1cd72d7333ce84975470f5d865383422034baa7d/include/fuse.h#L620
https://github.com/osxfuse/fuse/blob/abc35f85fdede186ed164082bd71fc1217d2f58f/lib/fuse.c#L4094

This change maps the Block Size to both upper-level block size and fragment size.

Previously it resulted into incorrect reporting of total and free space in the "df" command.

### DIFF
--- a/fuse/types_darwin.go
+++ b/fuse/types_darwin.go
@@ -148,6 +148,6 @@ func (s *StatfsOut) FromStatfsT(statfs *syscall.Statfs_t) {
 	s.Bavail = statfs.Bavail
 	s.Files = statfs.Files
 	s.Ffree = statfs.Ffree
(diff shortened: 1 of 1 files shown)


hanwen:
this needs a CLA.

the comments on the code you change suggest that I have actually thought about this when I wrote this, and that it wasn't a random oversight. 

Could you add a test that exercises statfs on OSX that exercises this new behavior?

hanwen:
no reponse for 6 months; closing.

PR #215, opened 2018-05-03
fuse: add unit-tests for loopback Utimens()

rfjakob:
Tests loopbackFileSystem.Utimens() and loopbackfile.Utimens()
at 1-second precision.

The exising TestUtimesNano() test only works on Linux
because it relies on syscall.UtimesNano(), which is not
available on Darwin. The new tests call the Utimens()
functions directly, bypassing FUSE, and work on all
platforms.

Because Darwin does not have syscall.UtimesNano(),
getting the Utimens() implementation right is hard.

The tests currently fail on Darwin, underlining the
need for them ( https://github.com/rfjakob/gocryptfs/issues/229 ):

$ go test ./fuse/nodefs
[...]
--- FAIL: TestLoopbackFileUtimens (0.00s)
	files_test.go:51: mtime has changed: 1525378914 -> 1073
	files_test.go:70: atime has changed: 1525291058 -> 1073
[...]

$ go test ./fuse/pathfs
--- FAIL: TestLoopbackFileSystemUtimens (0.00s)
	loopback_test.go:55: mtime has changed: 1525378929 -> 1073
	loopback_test.go:74: atime has changed: 1525291058 -> 1073
[...]

### DIFF
--- a/fuse/nodefs/files_test.go
+++ b/fuse/nodefs/files_test.go
@@ -1,0 +1,95 @@
+// Copyright 2016 the Go-FUSE Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package nodefs
+
+import (
+	"io/ioutil"
+	"os"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/hanwen/go-fuse/fuse"
+)
+
+// Check that loopbackFile.Utimens() works as expected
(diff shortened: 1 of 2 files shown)


rfjakob:
Hmm, it will be a PITA to keep TestLoopbackFileUtimens TestLoopbackFileSystemUtimens in sync, this needs to move into a shared helper function somehow. I'll resend later.

PR #217, opened 2018-05-08
fuse/nodefs: add LookupKnownChildren mount option

hanwen:
WIP/DONOTSUBMIT: should use LookupKnownChildren in all example code,
as it is the better default.


hanwen:
pushed directly.

PR #220, opened 2018-07-03
Updating kaoet's PR with your feedback

paulcwarren:
Kaoet's [PR](https://github.com/hanwen/go-fuse/pull/199) would be very useful for us.  However, he doesn't seem to have responded to your feedback so we are trying to pick up the thread.

We have asked questions on the original PR where we need some clarification.

Some other points to note:
- Unlike the original go-fuse code that is tested on Lucid.  We require to run on Trusty/Xenial and therefore tested on this platform.  We found that the --verbose option has been removed from flock in Trusty and therefore we had to relax the `TestFlockExclusive` test.  We werent too pleased with the result but see no other option.
- We added comments where you have asked.
- We added @kaoet 's name and email to the AUTHORS file.  Hopefully he wont mind!?!
- We added additional testing to ensure SetLk and SetLkw are called but we couldn't find a way to use flock to invoke GetLk.  We tried cgo in an attempt to call fcntl.c but cgo isnt allowed in test files.  Maybe you have some other ideas?

Feedback welcome.  We really would like to work with you to get this PR merged as it would make our lives easier :-)


hanwen:
Hi Paul, happy to oblige and merge whatever you produce, but I need a CLA signoff, from you or your employer.

Thanks!

kaoet:
To be honest, I implemented the flock just for my own project. And It just works fine these days.

Just pick up my code and do whatever you like. :)

hanwen:
verified CLA for Paul & Maria

hanwen:
There were several items on kaoets original PR that weren't addressed. Can you have a look at those?

Also, the test, if run with the race detector, triggers a data race. Can you add synchronization so it doesn't trigger the detector?

mariash:
@googlebot confirming

paulcwarren:
Thanks @hanwen.  Much appreciated.

Issue #223, opened 2018-07-25
Implement file locking: getlk/setlk/setlkw/flock


limbo127:
hello, 
this commit break the starting of qemu/libvirt when using go-fuse to mount a raw image, with 
Failed to lock byte 100   from qemu 

Regards, 
Nicolas Prochazka

hanwen:
fixed in 8393ebf1f6697608d77990be8139fcea1d271d3b

thanks for the report.

PR #225, opened 2018-07-27
Make it possible to remove the locking capabilities from server flags

folbricht:
Since https://github.com/hanwen/go-fuse/commit/1eca9006d0f2410c01fe65d69c536a36c146c460, fully implementing `GetLk`, `SetLk`, and `SetLkw` has become a requirement which breaks existing use-cases of this library, especially when it's not backed by an actual filesystem.

This PR is a first attempt at making this behavior configurable and allowing client implementations to disable these capabilities in fuse.Server.

If you believe there's a better way to achieve the same, please let me know and I'll update the PR accordingly.

This will address https://github.com/hanwen/go-fuse/issues/223

### DIFF
--- a/fuse/api.go
+++ b/fuse/api.go
@@ -74,6 +74,9 @@ type MountOptions struct {
 	// Xattr operations at all.
 	DisableXAttrs bool
 
+	// If set, disables the CAP_FLOCK_LOCKS and CAP_POSIX_LOCKS capabilities
+	DisableLocks bool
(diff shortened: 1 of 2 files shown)


hanwen:
I fixed this in https://github.com/hanwen/go-fuse/commit/8393ebf1f6697608d77990be8139fcea1d271d3b
and added a test to make sure it won't regress.

PR #232, opened 2018-09-22
pathfs: fix possible nil pointer dereference in GetAttr

rfjakob:
gocryptfs user Felix Lechner reported a nil pointer dereference
in GetAttr: https://github.com/rfjakob/gocryptfs/issues/260

The crash is in line

	n.setClientInode(fi.Ino)

because fi is nil.

This can happen when file.GetAttr() returns an error code other than
ENOSYS and EBADF. For gocryptfs, this can only happen when an open
file descriptor breaks. In this case it was triggered by a failing
NFS volume.

Fix the crash by erroring out for error codes that are not handled
by the retry-by-path logic.


lechner:
Hi, I tested this patch for some time. It fixed the issue. Please accept the pull request. Thank you!

PR #235, opened 2018-10-02
fuse: A bit better package documentation rendering in godoc

navytux:
Current state is: fuse's godoc package-level description is completely
empty because there is a blank link between package-level description
and package clause.

Yes, a user still has to go through api.go and read it in full, but
having at least something for the description is a better start, and
removing one line is easy.

We already do it this way e.g. for nodefs and in other places.


navytux:
Ping @hanwen. Any reason why we should keep this not merged?

Thanks beforehand,
Kirill

hanwen:
sorry, I didnt see it before.

navytux:
Thanks for merging.

PR #238, opened 2018-10-09
fuse: Fix build on 32-bit platforms

navytux:
@rfjakob reports that after bdca0e6a (Add support for store notify)
build got broken on ARM:

	$ GOARCH=arm go build ./example/loopback
	# github.com/hanwen/go-fuse/fuse
	fuse/server.go:490:11: constant 4294967295 overflows int
	fuse/server.go:492:9: constant 4294967295 overflows int

Fix it by amending "store notify" patch by chunking data into 2GB instead of
full-32bit-range 4GB.

### DIFF
--- a/fuse/server.go
+++ b/fuse/server.go
@@ -487,9 +487,11 @@ func (ms *Server) InodeNotifyStoreCache(node uint64, offset int64, data []byte) 
 
 	for len(data) > 0 {
 		size := len(data)
-		if size > math.MaxUint32 {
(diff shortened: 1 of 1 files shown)


navytux:
( travis failures are again in unionfs )

rfjakob:
Lgtm

navytux:
Thanks.

hanwen:
re: travis failures, see issue https://github.com/hanwen/go-fuse/issues/239

PR #240, opened 2018-10-10
fuse: Correct NOTIFY constants/operation names

navytux:
- it is not NOTIFY_INVAL_DELETE, but just NOTIFY_DELETE as the kernel is
  notified of entry being deleted and there is no invalidation here.
  uapi/linux/fuse.h does not use "_INVAL_" for NOTIFY_DELETE neither.

- similarly for "notify store/retrieve" _INVAL_ is not appropriate and
  neither is used in uapi/linux/fuse.h . However since we took more
  explicit approach for "notify store/retrieve" naming (see bdca0e6a
  "Add support for store notify") let's also add _CACHE suffix for
  "notify store/retrieve" operation names for consistency and for being
  less ambiguous.

- for inode/entry invalidation operations, let's use _INVAL_ in internal
  _OP_NOTIFY_*, similarly to how it is used in corresponding NOTIFY_*
  constant. For example:

	_OP_NOTIFY_ENTRY -> _OP_NOTIFY_INVAL_ENTRY	(corresponds to NOTIFY_INVAL_ENTRY)

### DIFF
--- a/fuse/misc.go
+++ b/fuse/misc.go
@@ -23,9 +23,9 @@ func (code Status) String() string {
 			"NOTIFY_POLL",
 			"NOTIFY_INVAL_INODE",
 			"NOTIFY_INVAL_ENTRY",
-			"NOTIFY_INVAL_STORE",
-			"NOTIFY_INVAL_RETRIEVE",
-			"NOTIFY_INVAL_DELETE",
+			"NOTIFY_STORE_CACHE",
+			"NOTIFY_RETRIEVE_CACHE",
+			"NOTIFY_DELETE",
 		}[-code]
 	}
 	return fmt.Sprintf("%d=%v", int(code), syscall.Errno(code))
--- a/fuse/opcode.go
+++ b/fuse/opcode.go
@@ -58,10 +58,10 @@ const (
 	_OP_FUSE_RENAME2 = int32(45) // protocol version 23.
 
 	// The following entries don't have to be compatible across Go-FUSE versions.
-	_OP_NOTIFY_ENTRY  = int32(100)
-	_OP_NOTIFY_INODE  = int32(101)
-	_OP_NOTIFY_STORE  = int32(102)
-	_OP_NOTIFY_DELETE = int32(103) // protocol version 18
+	_OP_NOTIFY_INVAL_ENTRY = int32(100)
+	_OP_NOTIFY_INVAL_INODE = int32(101)
+	_OP_NOTIFY_STORE_CACHE = int32(102)
+	_OP_NOTIFY_DELETE      = int32(103) // protocol version 18
 
 	_OPCODE_COUNT = int32(104)
 )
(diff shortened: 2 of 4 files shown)


navytux:
( @hanwen, for CI Go1.9 succeeded, but Go1.10, Go1.11 and Go tip failed in unionfs again )

fuse/misc.go
```diff
@@ -23,9 +23,9 @@ func (code Status) String() string {
 			"NOTIFY_POLL",
 			"NOTIFY_INVAL_INODE",
```
hanwen: it never occurred to me that "inval" stands for invalidate. Maybe add a comment somewhere.

navytux: Yes, FUSE protocol is not very explicitly documented, but it could be seen that `_INVAL_` indeed stands for invalidations:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/fuse.h?id=v4.19-rc6-177-gcec4de302c5f#n58
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/fuse.h?id=v4.19-rc6-177-gcec4de302c5f#n389

Probably something as below would do:

```diff
--- a/fuse/types.go
+++ b/fuse/types.go
@@ -384,12 +384,12 @@ type NotifyStoreOut struct {
 }
 
 const (
-       //      NOTIFY_POLL         = -1
-       NOTIFY_INVAL_INODE = -2
-       NOTIFY_INVAL_ENTRY = -3
-       NOTIFY_STORE_CACHE = -4 // "notify store" in uapi/linux/fuse.h
-       //      NOTIFY_RETRIEVE_CACHE  = -5 // "notify retrieve" in uapi/linux/fuse.h
-       NOTIFY_DELETE      = -6
+       //      NOTIFY_POLL         = -1 // notify kernel that a poll waiting for IO on a file handle should wake up
+       NOTIFY_INVAL_INODE = -2 // notify kernel that an inode should be invalidated
+       NOTIFY_INVAL_ENTRY = -3 // notify kernel that a directory entry should be invalidated
+       NOTIFY_STORE_CACHE = -4 // store data into kernel cache of an inode
+       //      NOTIFY_RETRIEVE_CACHE  = -5 // retrieve data from kernel cache of an inode
+       NOTIFY_DELETE      = -6 // notify kernel that a directory entry has been deleted
 
 //     NOTIFY_CODE_MAX     = -6
 )
```

?

Also on this topic: I find that e.g. `fuse.Server.InodeNotify()` / / `nodefs.FileSystemConnector.FileNotify()` naming is a bit misleading. I would suggest to consider changing e.g. to `Server.NotifyInodeInvalidate()` and `FileSystemConnector.NotifyFileInvalidate()` but I'm not sure we can change public API.

( for the reference full names with such approach e.g. for FileSystemConnector:

- `NotifyFileInvalidate()`
- `NotifyEntryInvalidate()`
- `NotifyFileStoreCache()`
- `FileRetrieveCache()`   // not only it notifies, but also waits for data back from the kernel
- `NotifyEntryDelete()`

)

navytux:
( I have updated the patch with ^^^ comments to `NOTIFY_*` constants )

navytux:
( CI failures are as before )

fuse/misc.go
hanwen: we could change the name, but it seems a small improvement with a large cost.

If we do anything, there should probably be much more docs on how to use the library.

navytux: I agree. Let's keep this in mind if we ever get to there.

navytux:
Thanks for merging.

PR #241, opened 2018-10-12
fuse: Package-level overview

navytux:
I've made the notes for myself, but then though it could be maybe useful
as the package documentation.

### DIFF
--- a/fuse/api.go
+++ b/fuse/api.go
@@ -2,9 +2,86 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// The fuse package provides APIs to implement filesystems in
-// userspace.  Typically, each call of the API happens in its own
+// Package fuse provides APIs to implement filesystems in
+// userspace in terms of raw FUSE protocol.
+//
+// A filesystem is implemented by implementing its server that provides a
+// RawFileSystem interface. Typically the server embeds
+// NewDefaultRawFileSystem() and implements only subset of filesystem methods:
+//
+//	type MyFS struct {
+//		fuse.RawFileSystem
+//		...
+//	}
+//
+//	func NewMyFS() *MyFS {
+//		return &MyFS{
+//			RawFileSystem: fuse.NewDefaultRawFileSystem(),
+//			...
+//		}
+//	}
+//
+//	// Mkdir implements "mkdir" request handler.
+//	//
+//	// For other requests - not explicitly implemented by MyFS - ENOSYS
+//	// will be typically returned to client.
+//	func (fs *MyFS) Mkdir(...) {
+//		...
+//	}
+//
+// Then the filesystem can be mounted and served to a client (typically OS
+// kernel) by creating Server:
+//
+//	fs := NewMyFS() // implements RawFileSystem
+//	fssrv, err := fuse.NewServer(fs, mountpoint, &fuse.MountOptions{...})
+//	if err != nil {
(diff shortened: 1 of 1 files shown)


navytux:
( CI is broken in unionfs as before - no regression here as this is documentation only change )

fuse/api.go
```diff
@@ -2,9 +2,74 @@
...
+// kernel) by creating Server:
+//
+//	fs := MyFSNew() // implements RawFileSystem
+//	fssrv, err := fuse.NewServer(fs, mountpoint, &fuse.MountOptions{...})
+//	if err != nil {
+//		...
+//	}
+//
+//	fssrv.Serve()
+//
```
hanwen: there should be a call for WaitMount() somewhere, if the server wants to interact with the FS too, like in tests.

fuse/api.go
```diff
@@ -2,9 +2,74 @@
...
+// implement filesystem at higher levels:
+//
+// Package github.com/hanwen/go-fuse/fuse/nodefs provides way to implement
+// filesystems in terms of inodes. This resembles kernel's idea of what a
+// filesystem looks like.
+//
+// Package github.com/hanwen/go-fuse/fuse/pathfs provides way to implement
+// filesystems in terms of path names. Working with path names is somewhat
+// easier compared to inodes, however do not use pathfs if you care about
+// correctness.
```
hanwen: this needs some more qualification. The tricky bit is what to do with operations that change the name => inode mapping in combination with open files. It can be done correctly, but if using nodefs, you get it for free.

rfjakob:
Is this "do not use pathfs if you care about correctness" still a thing? Pathfs passes xfstests just fine these days

hanwen:
@rfjakob - we still have a few failures, no?

The thing that worries me is described in the thread

https://sourceforge.net/p/fuse/mailman/fuse-devel/thread/BANLkTik78We10iL%3DYCdCuSQe4U-R_2K2NA%40mail.gmail.com/#msg27550667

go-fuse doesnt have the custom r/w lock so I think it is vulnerable to races.

rfjakob:
True, but nothing related to renames: https://github.com/rfjakob/gocryptfs/blob/master/Documentation/XFSTESTS.md . We used to break at fsstress, but this is long gone

I'll try to repro the problem from the thread

fuse/api.go
navytux: Agree. I did not put it in there becuase my filesystem runs .Serve() synchronously, but you are right. Reworking this to:

```diff
 //     if err != nil {
 //             ...
 //     }
 //
+// and letting the server do its work:
+//
+//     // either synchronously - .Serve() blocks until the filesystem is unmounted.
 //     fssrv.Serve()
 //
+//     // or in the background - .Serve() is spawned in another goroutine, but
+//     // before interacting with fssrv from current context we have to wait
+//     // until the filesystem mounting is complete.
+//     go fssrv.Serve()
+//     err = fssrv.WaitMount()
+//     if err != nil {
+//             ...
+//     }
+//
```

By the way - `.Serve()` does not return any error - is that good? In my view `.Serve()` should always return an error, and the error returned should indicate the reason Serve finished - either e.g. filesystem was unmounted normally, or another e.g. IO error with the kernel via /dev/fuse or whatever else.

Maybe we can add the error return to `.Serve()` because it does not break compatibility?

navytux:
Thanks for merging.

PR #242, opened 2018-10-26
added syscall.EISDIR constant

Muryoutaisuu:
for convenience reasons like return values of `fuse.Status` in type FileSystem

e.g. in `Open(name string, flags uint32, context *fuse.Context) (file nodefs.File, code fuse.Status)`


hanwen:
this looks OK, but I'd expect the VFS to never call open on something it knows is a directory.

Muryoutaisuu:
@hanwen  thanks for merging. It's rather for error handling when users call a cat on a directory -> so they get a consistent error message

PR #247, opened 2018-12-11
added EREMOTEIO constant

muryoutaisuu:
my fuse implementation connects to another backend. I want users to be able to tell, whether a IO error happens on my fuse implementation itself (because users did some wrong interaction or sth. else) or rather on the backend (because backend itself returns an EIO error, which would be wrongy interpreted if my implementation returned EIO as well).

It's not a MUST, I could still live with the EIO option. But would be very nice.

### DIFF
--- a/fuse/types.go
+++ b/fuse/types.go
@@ -34,6 +34,9 @@ const (
 	// EIO I/O error
 	EIO = Status(syscall.EIO)
 
+	// EREMOTEIO Remote I/O error
+	EREMOTEIO = Status(syscall.EREMOTEIO)
+
 	// ENOENT No such file or directory
 	ENOENT = Status(syscall.ENOENT)
 


hanwen:
thanks!

rfjakob:
Unfortunately, this constant does not exist on macos: 

+GOOS=darwin
+GOARCH=amd64
+go build -tags without_openssl
# github.com/hanwen/go-fuse/fuse
../../hanwen/go-fuse/fuse/types.go:38:21: undefined: syscall.EREMOTEIO

navytux:
For the reference - for my filesystem I gave up on the idea to expose all possible error codes via fuse package and just convert `syscall.Errno` to `fuse.Status` as needed:

```go
// err2LogStatus converts an error into FUSE status code and logs it appropriately.
//
// the error is logged because otherwise, if e.g. returning EINVAL or EIO
// codes, there is no more detail except the error code itself.
func err2LogStatus(err error) fuse.Status {
        // no error
        if err == nil {
                return fuse.OK
        }

        // direct usage of error code - don't log
        ecode, iscode := err.(syscall.Errno)
        if iscode {
                return fuse.Status(ecode)
        }

        // otherwise log as warnings EINVAL and as errors everything else
        switch errors.Cause(err).(type) {
        case *eInvalError:
                log.Warning(err)
                return fuse.EINVAL

        default:
                log.Error(err)
                return fuse.EIO
        }
}

...

// eInvalError is the error wrapper signifying that underlying error is about "invalid argument".
// err2LogStatus converts such errors into EINVAL return code + logs as warning.
type eInvalError struct {
        err error
}

func (e *eInvalError) Error() string {
        return "invalid argument: " + e.err.Error()
}

// don't propagate eInvalError.Cause -> e.err

func eINVAL(err error) *eInvalError {
        return &eInvalError{err}
}

func eINVALf(format string, argv ...interface{}) *eInvalError {
        return eINVAL(fmt.Errorf(format, argv...))
}

...

// /bigfile -> Mkdir receives client request to create /bigfile/<bigfileX>.
//
// It creates <bigfileX>/head/* along the way.
func (bfroot *BigFileRoot) Mkdir(name string, mode uint32, fctx *fuse.Context) (*nodefs.Inode, fuse.Status) {
        inode, err := bfroot.mkdir(name, fctx) // XXX ok to ignore mode?
        return inode, err2LogStatus(err)

}
```

muryoutaisuu:
@rfjakob thanks for the hint. Didn't know it does not exist on macOS.
@navytux thanks for the workaround, I'll do it as you proposed.

I just created @248 to revert the changes.

navytux:
@muryoutaisuu, you are welcome.

rfjakob:
I don't think we have to revert this - instead we could define it to be identical to EIO on macos

muryoutaisuu:
yeah I just saw, that @hanwen already moved it into linux types with this commit: 8f0f5c3f125ff4b698108fc986d45b931c487722

rfjakob:
Ok perfect!

PR #249, opened 2018-12-12
fuse: ↑ signal/noise in log messages

navytux:
- use tA=...s instead of `A...` or `AttrValid=...` for time(Attr valid);
- use tE=...s instead of `EntryValid` for time(Entry valid);
- use iX instead of `NodeId: X` for inode X;
- use gX instead of `Generation=X` for generation X;
- use `[<off> +<size>)` instead of `off <off> sz <sz>`;

- use

	< uniq: iX.Op opargs
	> uniq: status reply...

  instead of

	Dispatch uniq: Op NodeId: X opargs
	Serialize uniq: Op code: status reply...

  for received/replied messages;

- use `["file1", "file2"]` instead of `names: [file1, file2]` for filenames.

- use `xb` instead of `x bytes`

- for replied data, also log data content, but only up to first 8 bytes.
  This does not flood the log, but at the same time makes filesystem
  diagnostics easier.

  If this last change is controversial, it could be dropped from the
  patch.

Overall effect. Please compare how it was before:

	wcfs: 17:09:21.910510 Dispatch 1: INIT, NodeId: 0. data: {7.27 Ra 0x20000 ASYNC_READ,SPLICE_WRITE,FLOCK_LOCKS,IOCTL_DIR,READDIRPLUS_AUTO,ASYNC_DIO,CAP_POSIX_ACL,POSIX_LOCKS,DONT_MASK,SPLICE_MOVE,AUTO_INVAL_DATA,CAP_PARALLEL_DIROPS,ATOMIC_O_TRUNC,SPLICE_READ,NO_OPEN_SUPPORT,CAP_PARALLEL_DIROPS,EXPORT_SUPPORT,BIG_WRITES,READDIRPLUS,WRITEBACK_CACHE,0x200000}
	wcfs: 17:09:21.911719 Serialize 1: INIT code: OK value: {7.23 Ra 0x20000 NO_OPEN_SUPPORT,BIG_WRITES,READDIRPLUS,ASYNC_READ,AUTO_INVAL_DATA 0/0 Wr 0x10000 Tg 0x0}
	wcfs: 17:09:22.002776 Dispatch 2: LOOKUP, NodeId: 1. names: [.wcfs] 6 bytes
[...]

PR #251, opened 2019-01-11
readme: Add appendix about how to read Go-FUSE debug log output

navytux:
d1c826d1 ("fuse: increase signal/noise in log messages",
https://github.com/hanwen/go-fuse/pull/249) changed log to use
abbreviations in the name of improving signal/noise ratio. However it
might be not evident offhand what an abbreviation means.

Add a short reference in the readme with the summary on how to read the
log and a short example.

The summary is incomplete - for example I don't describe Fh and other
attributes that d1c826d1 did not change. I propose we review logging
more, e.g. consider doing `Fh X` -> `fX` and maybe similarly for other
fields, and while doing so, also populate/update the summary and example
in the readme. In other words it is good to document current state, but
it is not fixed in stone and can be evolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,3 +147,45 @@ Grep source code for TODO.  Major topics:
 
 Like Go, this library is distributed under the new BSD license.  See
 accompanying LICENSE file.
+
+--------
+
+## Appendix I. Go-FUSE log format
+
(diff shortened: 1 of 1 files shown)


navytux:
( CI is flaky in unionfs and the failure is not related to readme update at all )

navytux:
Thanks for merging the patch.

Issue #252, opened 2019-01-17
ReadDir returns outdated directory content


slackner:
First of all, thanks for this awesome project! I'm currently debugging an issue I found with gocryptfs (which uses go-fuse) on macOS, and I hope its okay to open a bug report here to get some more opinions involved. The bug is https://github.com/rfjakob/gocryptfs/issues/351, however, I will provide a short summary/analysis below.

The issue occurs in a test that repeatedly creates files and queries directory listings with multiple FUSE systems mounted on top of each other. While this test works fine on Linux, on macOS apparently at some point the directory listings contain outdated information. The exact timing of the failure is more or less random. Looking at the debug output shows that, at the point where the application calls `OpenDir`, there is no corresponding FUSE call. Instead, the following `ReadDir` calls use an existing file handle?!

Relevant log (shortened and annotated):

```
2019/01/15 12:02:45 rx 2: OPENDIR i1 
2019/01/15 12:02:45 tx 2:     OK, {Fh 2 }

    ... file is created here ...

2019/01/15 12:02:45 rx 4: LOOKUP i1 ["gocryptfs.longname.iNPx_rqvgCZCkSHiDxThVRUMDqfqDYPSMcu58uAVTdA"] 63b
2019/01/15 12:02:45 tx 4:     OK, {i330 g329 tE=1s tA=1s {M0100644 SZ=0 L=1 503:204 0 0:3182331 A 1547582565.000000 M 1547582565.000000 C 1547582565.000000}}
[...]
2019/01/15 12:02:45 rx 4: OPEN i330 {O_RDONLY} 
2019/01/15 12:02:45 tx 4:     OK, {Fh 3 }
[...]
2019/01/15 12:02:45 rx 2: RELEASE i330 {Fh 3   L0} 
2019/01/15 12:02:45 tx 2:     OK

    ... opening the file seems to work fine ...
    ... here is an OpenDir call in the source, but there is no matching OPENDIR fuse call ...

2019/01/15 12:02:45 rx 3: READDIR i1 {Fh 2 [0 +6553) } 
2019/01/15 12:02:45 tx 3:     OK,  328b data "\u038d0\x00\x00\x00\x00\x00"...
2019/01/15 12:02:45 rx 2: READDIR i1 {Fh 2 [6 +6553) } 
2019/01/15 12:02:45 tx 2:     OK

    ... unsurprisingly the application gets the outdated directory listing ...
```

[...]

PR #254, opened 2019-01-24
nodefs: delay directory read till ReadDir

rfjakob:
We used to read the directory already on OpenDir, which means that
we missed changes in the directory between a user's open() and
readdir() call.

https://github.com/hanwen/go-fuse/issues/252

### DIFF
--- a/fuse/nodefs/dir.go
+++ b/fuse/nodefs/dir.go
@@ -13,33 +13,30 @@ import (
 
 type connectorDir struct {
 	node  Node
+	inode *Inode
 	rawFS fuse.RawFileSystem
 
 	// Protect stream and lastOffset.  These are written in case
 	// there is a seek on the directory.
 	mu     sync.Mutex
 	stream []fuse.DirEntry
-
-	// lastOffset stores the last offset for a readdir. This lets
-	// readdir pick up changes to the directory made after opening
-	// it.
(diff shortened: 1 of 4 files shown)


navytux:
@rfjakob, could we please have a unit test for this fix?

rfjakob:
Agreed. Will add.

fuse/nodefs/dir.go
```diff
@@ -13,33 +13,30 @@ import (
 
 type connectorDir struct {
 	node  Node
+	inode *Inode
 	rawFS fuse.RawFileSystem
 
 	// Protect stream and lastOffset.  These are written in case
```
slackner: There is also a remaining reference to `lastOffset`. We might want to update that comment aswell.

hanwen:
if we do this, OpenDir failures will manifest on readdir() rather than open(). Is that a problem?

rfjakob:
Thing is, the kernel issues a lookup before opendir, so this will happen rareley.

rfjakob:
Tests added! (travis failures are from unionfs)

navytux:
@rfjakob, thanks for the test.

rfjakob:
@hanwen done

hanwen:
thanks, I rebased and pushed.

PR #257, opened 2019-01-26
Don't crash when calling truncate(1) on a handle without a File.

EdSchouten:
For all operations that can be applied on file descriptors, we have
special treatment to permit the File to be nil. In that case, operations
are called against the Node object instead.

This check currently seems missing from SetAttr(), causing invocations
of the truncate(1) command line tool to cause a crash:

	panic: runtime error: invalid memory address or nil pointer dereference
	[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xa30d0d]

	goroutine 12 [running]:
	github.com/hanwen/go-fuse/fuse/nodefs.(*rawBridge).SetAttr(0xc00013ecc0, 0xc0003a8000, 0xc0003ba0f0, 0xa27b98)
		external/com_github_hanwen_go_fuse/fuse/nodefs/fsops.go:203 +0x58d
	github.com/hanwen/go-fuse/fuse.doSetattr(0xc0003a4000, 0xc0003ba000)
		external/com_github_hanwen_go_fuse/fuse/opcode.go:171 +0x64
	github.com/hanwen/go-fuse/fuse.(*Server).handleRequest(0xc0003a4000, 0xc0003ba000, 0xc0003ba000)
		external/com_github_hanwen_go_fuse/fuse/server.go:431 +0x26b
	github.com/hanwen/go-fuse/fuse.(*Server).loop(0xc0003a4000, 0xc00000b101)
		external/com_github_hanwen_go_fuse/fuse/server.go:403 +0x18f
	created by github.com/hanwen/go-fuse/fuse.(*Server).readRequest
		external/com_github_hanwen_go_fuse/fuse/server.go:291 +0x2d8

### DIFF
--- a/fuse/nodefs/fsops.go
+++ b/fuse/nodefs/fsops.go
@@ -199,8 +199,9 @@ func (c *rawBridge) SetAttr(input *fuse.SetAttrIn, out *fuse.AttrOut) (code fuse
 
 	var f File
 	if input.Valid&fuse.FATTR_FH != 0 {
-		opened := node.mount.getOpenedFile(input.Fh)
-		f = opened.WithFlags.File
+		if opened := node.mount.getOpenedFile(input.Fh); opened != nil {
+			f = opened.WithFlags.File
+		}
 	}
 
 	if code.Ok() && input.Valid&fuse.FATTR_MODE != 0 {
--- a/fuse/test/nil_file_truncation_test.go
+++ b/fuse/test/nil_file_truncation_test.go
@@ -1,0 +1,77 @@
+// Copyright 2019 the Go-FUSE Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package test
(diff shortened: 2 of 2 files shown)


navytux:
Could you please add a test? By the way, does it still crash without your patch, buf if https://github.com/hanwen/go-fuse/pull/258 is applied?

EdSchouten:
Hi there! It still crashes with #258 applied. I'll add test case.

navytux:
Hi, thanks for feedback and the test. For me the patch looks ok.

Grepping for `getOpenedFile` I see we have similar `nil` dereference errors in `ReadDir` and `ReadDirPlus`, at least if the kernel sends invalid input.

navytux:
@hanwen, is there any reason we don't merge this bugfix? If it is only about test - I could integrate it into `nofile_test.go` after the merge.

hanwen:
sorry, done.

Issue #168, opened 2017-07-15
TestFopenKeepCache is flaky.


hanwen:
ok  	github.com/hanwen/go-fuse/fuse/test	1.948s
--- FAIL: TestFopenKeepCache (0.03s)
	cache_test.go:108: ReadFile: open /tmp/TestFopenKeepCache983114099/mnt/file.txt: no such file or directory

navytux:
This is still true of today's master (32b02fb):

```
.../github.com/hanwen/go-fuse/fuse/test$ while true; do go test -count 1 .; done
ok      github.com/hanwen/go-fuse/fuse/test     2.368s
ok      github.com/hanwen/go-fuse/fuse/test     2.619s
--- FAIL: TestFopenKeepCache (0.03s)
    cache_test.go:108: ReadFile: open /tmp/TestFopenKeepCache932305376/mnt/file.txt: no such file or directory
11:17:00.327018 Unimplemented opcode INTERRUPT
11:17:00.327249 writer: Write/Writev failed, err: 2=no such file or directory. opcode: INTERRUPT
FAIL
FAIL    github.com/hanwen/go-fuse/fuse/test     2.522s
ok      github.com/hanwen/go-fuse/fuse/test     2.646s
ok      github.com/hanwen/go-fuse/fuse/test     2.784s
ok      github.com/hanwen/go-fuse/fuse/test     2.788s
ok      github.com/hanwen/go-fuse/fuse/test     2.597s
ok      github.com/hanwen/go-fuse/fuse/test     2.869s
ok      github.com/hanwen/go-fuse/fuse/test     2.767s
ok      github.com/hanwen/go-fuse/fuse/test     2.692s
ok      github.com/hanwen/go-fuse/fuse/test     2.878s
--- FAIL: TestFopenKeepCache (0.04s)
    cache_test.go:108: ReadFile: open /tmp/TestFopenKeepCache712352879/mnt/file.txt: no such file or directory
11:17:28.414978 Unimplemented opcode INTERRUPT
11:17:28.415085 writer: Write/Writev failed, err: 2=no such file or directory. opcode: INTERRUPT
FAIL
FAIL    github.com/hanwen/go-fuse/fuse/test     2.838s
ok      github.com/hanwen/go-fuse/fuse/test     2.353s
ok      github.com/hanwen/go-fuse/fuse/test     2.676s
ok      github.com/hanwen/go-fuse/fuse/test     2.735s
[...]

PR #262, opened 2019-02-03
Prevent concurrent LOOKUP and FORGET

rfjakob:
Add a lock that prevents LOOKUP and FORGET from running concurrently.
    
Locking at this level is a big hammer, but makes sure we don't return
forgotten nodes to the kernel. Problems solved by this lock:
    
https://github.com/hanwen/go-fuse/issues/168
https://github.com/rfjakob/gocryptfs/issues/322

### DIFF
--- a/fuse/nodefs/fsconnector.go
+++ b/fuse/nodefs/fsconnector.go
@@ -12,6 +12,7 @@ import (
 	"log"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 	"unsafe"
 
@@ -35,6 +36,13 @@ type FileSystemConnector struct {
 
 	// The root of the FUSE file system.
 	rootNode *Inode
+
+	// This lock prevents Lookup() and Forget() from running concurrently.
+	// Locking at this level is a big hammer, but makes sure we don't return
+	// forgotten nodes to the kernel. Problems solved by this lock:
+	// https://github.com/hanwen/go-fuse/issues/168
+	// https://github.com/rfjakob/gocryptfs/issues/322
+	lookupLock sync.Mutex
 }
 
 // NewOptions generates FUSE options that correspond to libfuse's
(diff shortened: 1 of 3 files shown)


rfjakob:
The patch `pathfs: add nodeid to ".deleted" placeholder` helped debugging the issue, but does not matter in terms of correctness.

rfjakob:
xfstests looks good: https://gist.github.com/rfjakob/0537715c1264a6f494680e77d83f3882. As expected, generic/273 is fixed by this change, and it does not introduce new failures.

hanwen:
it's a mess. I think this will solve it, but the more I think of it, the more I want to write a completely new nodefs API that changes the directory tree and lookup counts under the same lock.

hanwen:
what about readdirplus ?

rfjakob:
W.r.t. a rewrite: I also thought about it a short time. Merging pathfs and nodefs would simplify things a little, but it's still a big task. I'd certainly be motivated to help, though, and also to port gocryptfs over if the API changes.

hanwen:
Part of me think that go-fuse also had its best time; there is no support for contexts, for example, and the in-process submounts were a mistake.

I've been eyeing https://github.com/jacobsa/fuse which looks nice

It seems most other packages don't make an effort to help the user in processing FORGET opcodes in the correct way.

re: pathfs: since we keep track of parent pointers, we can move GetPath into nodefs, mostly obviating pathfs.

navytux:
For the reference: before starting my filesystem I was too looking into jacobsa/fuse. That project is indeed better structured and better documented, but out of the box for me it worked ~ 1.6 times slower compared to hanwen/go-fuse:

https://lab.nexedi.com/kirr/go-fuse/commit/0aff2383d1

I did not investigate what was the reason, since hanwen/go-fuse also had better FUSE protocol coverage at that time.

hanwen:
I opened https://github.com/hanwen/go-fuse/issues/264 for discussing a new API.

PR #263, opened 2019-02-04
nodefs: Drop unneecessary casts

navytux:
- casting fuse.AttrOut.Attr to fuse.Attr is noop
- casting fileSystemMount.options.Owner to fuse.Owner is noop
- casting fuse.ReadIn.Context to fuse.Context is noop


navytux:
Thanks.

Issue #137, opened 2016-10-28
Regression: xfstests generic/035 fails because Nlink is forced to 1


rfjakob:
This seems to have been introduced in 3e9dd8f9e17171580aa481e9bee415a8546535d7 and causes xfstests generic/035 to fail with this:

```
 QA output created by 035
 overwriting regular file:
+nlink is 1, should be 0
 overwriting directory:
```

I suggest a revert.

rfjakob:
What this test does is
1) create two files, `a` and `b`
2) open `b`
3) rename `a` over `b`
4) fstat `b` and check nlink (should be zero)

hanwen:
I don't think you could ever get that test to pass due to https://github.com/libfuse/libfuse/issues/62 - the kernel doesn't issue a FH for GetAttr calls.

hanwen:
Correction: this test can pass if you build an in-memory node based FS. For loopback, you don't get FH, so there is no path to stat.

rfjakob:
I see your point, and I also remember adding that comment in 8b95cbfda9742b2781cb8481fd04fa2101762529 .

However, xfstests generic/035 at the moment looks like this:
```
$ cat fuse-xfstests/results/generic/035.out.bad
QA output created by 035
overwriting regular file:
nlink is 1, should be 0
overwriting directory:
t_rename_overwrite: fstat(3): No such file or directory
```

And with 3e9dd8f9e17171580aa481e9bee415a8546535d7 reverted it looks like this:
```
$ cat fuse-xfstests/results//generic/035.out.bad
QA output created by 035
overwriting regular file:
overwriting directory:
t_rename_overwrite: fstat(3): No such file or directory
```

I'll have to check the xfstests code. Maybe it actually gets an error from the fstat but does not report it.

rfjakob:
I'm not sure how, but it looks like the fstat succeeds: [code](https://github.com/rfjakob/fuse-xfstests/blob/gocryptfs-2018-08-18/src/t_rename_overwrite.c#L31)

I have added a debug printf [(commit)](https://github.com/rfjakob/fuse-xfstests/commit/18d8c59f3a4da81bbd00eabbed67e6d030823344), and the fstat result looks sane:
```
QA output created by 035
overwriting regular file:
res=0 dev=75 ino=9180181 mode=100644 nlink=0 uid=0
overwriting directory:
t_rename_overwrite: fstat(3): No such file or directory
[...]

PR #265, opened 2019-02-05
Newapi

hanwen:
sketch for a new nodefs API. What do you think?

@rfjakob 
@navytux

### DIFF
--- a/nodefs/bridge.go
+++ b/nodefs/bridge.go
@@ -29,6 +29,8 @@ type rawBridge struct {
 	options Options
 	root    *Inode
 
+	// mu protects the following data.  Locks for inodes must be
+	// taken before rawBridge.mu
 	mu    sync.Mutex
 	nodes []mapEntry
 	free  []uint64
@@ -89,16 +91,14 @@ func (b *rawBridge) Lookup(header *fuse.InHeader, name string, out *fuse.EntryOu
 
 	lockNodes(parent, child)
 	parent.setEntry(name, child)
-	unlockNodes(parent, child)
-
 	b.mu.Lock()
 	if child.nodeID == 0 {
 		b.registerInode(child)
 	}
 	out.NodeId = child.nodeID
-	b.mu.Unlock()
-
 	out.Generation = b.nodes[out.NodeId].generation
+	b.mu.Unlock()
+	unlockNodes(parent, child)
 
 	if b.options.AttrTimeout != nil {
 		out.SetAttrTimeout(*b.options.AttrTimeout)
@@ -110,7 +110,8 @@ func (b *rawBridge) Lookup(header *fuse.InHeader, name string, out *fuse.EntryOu
 	return fuse.OK
 }
 
-// registerInode sets an inode number in the child. Must have bridge.mu
+// registerInode sets an nodeID in the child. Must have bridge.mu and
+// child.mu
 func (b *rawBridge) registerInode(child *Inode) {
 	if l := len(b.free); l > 0 {
 		last := b.free[l-1]
@@ -141,17 +142,15 @@ func (b *rawBridge) Create(input *fuse.CreateIn, name string, out *fuse.CreateOu
 
 	lockNodes(parent, child)
 	parent.setEntry(name, child)
-	unlockNodes(parent, child)
-
 	b.mu.Lock()
 	if child.nodeID == 0 {
 		b.registerInode(child)
 	}
 	out.Fh = b.registerFile(f)
-	b.mu.Unlock()
-
 	out.NodeId = child.nodeID
 	out.Generation = b.nodes[child.nodeID].generation
+	b.mu.Unlock()
+	unlockNodes(parent, child)
 
 	if b.options.AttrTimeout != nil {
 		out.SetAttrTimeout(*b.options.AttrTimeout)
@@ -171,10 +170,12 @@ func (b *rawBridge) Forget(nodeid, nlookup uint64) {
 	n := b.nodes[nodeid].inode
 	b.mu.Unlock()
 
-	if forgotten, _ := n.removeRef(nlookup, false); forgotten {
-		b.free = append(b.free, nodeid)
-		b.nodes[nodeid].inode = nil
-	}
+	n.removeRef(nlookup, false)
+}
+
+func (b *rawBridge) unregisterNode(nodeid uint64) {
+	b.free = append(b.free, nodeid)
+	b.nodes[nodeid].inode = nil
 }
 
 func (b *rawBridge) SetDebug(debug bool) {}
--- a/nodefs/bridge.go
+++ b/nodefs/bridge.go
@@ -100,14 +100,17 @@ func (b *rawBridge) Lookup(header *fuse.InHeader, name string, out *fuse.EntryOu
 	b.mu.Unlock()
 	unlockNodes(parent, child)
 
+	b.setEntryOutTimeout(out)
+	return fuse.OK
+}
+
+func (b *rawBridge) setEntryOutTimeout(out *fuse.EntryOut) {
 	if b.options.AttrTimeout != nil {
 		out.SetAttrTimeout(*b.options.AttrTimeout)
 	}
 	if b.options.EntryTimeout != nil {
 		out.SetEntryTimeout(*b.options.EntryTimeout)
 	}
-
-	return fuse.OK
 }
 
 // registerInode sets an nodeID in the child. Must have bridge.mu and
@@ -152,12 +155,7 @@ func (b *rawBridge) Create(input *fuse.CreateIn, name string, out *fuse.CreateOu
 	b.mu.Unlock()
 	unlockNode2(parent, child)
 
-	if b.options.AttrTimeout != nil {
-		out.SetAttrTimeout(*b.options.AttrTimeout)
-	}
-	if b.options.EntryTimeout != nil {
-		out.SetEntryTimeout(*b.options.EntryTimeout)
-	}
+	b.setEntryOutTimeout(out)
 
 	out.OpenFlags = flags
 
--- a/nodefs/bridge.go
+++ b/nodefs/bridge.go
@@ -1,0 +1,439 @@
+// Copyright 2019 the Go-FUSE Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package nodefs
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/hanwen/go-fuse/fuse"
+)
+
+type mapEntry struct {
+	generation uint64
+	inode      *Inode
+}
+
+type fileEntry struct {
+	file File
+
+	// space to hold directory stuff
+}
+
+type rawBridge struct {
+	fuse.RawFileSystem
+
+	options Options
+	root    *Inode
+
+	mu    sync.Mutex
+	nodes []mapEntry
+	free  []uint64
+
+	files     []fileEntry
+	freeFiles []uint64
+}
+
+func NewNodeFS(root Node, opts *Options) fuse.RawFileSystem {
+	bridge := &rawBridge{
+		RawFileSystem: fuse.NewDefaultRawFileSystem(),
+	}
+
+	if opts != nil {
+		bridge.options = *opts
+	} else {
+		oneSec := time.Second
+		bridge.options.EntryTimeout = &oneSec
+		bridge.options.AttrTimeout = &oneSec
+	}
+
+	bridge.root = &Inode{
+		nodeID:      1,
+		lookupCount: 1,
+		mode:        fuse.S_IFDIR,
+		children:    make(map[string]*Inode),
+		parents:     nil,
+		node:        root,
+		bridge:      bridge,
+	}
+	root.setInode(bridge.root)
+	bridge.nodes = append(bridge.nodes,
+		mapEntry{},
+		// ID 1 is always the root.
+		mapEntry{inode: bridge.root})
+
+	// Fh 0 means no file handle.
+	bridge.files = []fileEntry{{}}
+	return bridge
+}
+
+func (b *rawBridge) inode(id uint64, fh uint64) (*Inode, fileEntry) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.nodes[id].inode, b.files[fh]
+}
+
+func (b *rawBridge) Lookup(header *fuse.InHeader, name string, out *fuse.EntryOut) (status fuse.Status) {
+	parent, _ := b.inode(header.NodeId, 0)
+
+	child, code := parent.node.Lookup(context.TODO(), name, out)
+	if !code.Ok() {
+		if b.options.NegativeTimeout != nil {
+			out.SetEntryTimeout(*b.options.NegativeTimeout)
+		}
+		return code
+	}
+
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	lockNodes(parent, child)
+	parent.setEntry(name, child)
+	unlockNodes(parent, child)
+
+	if child.nodeID == 0 {
+		b.registerInode(child)
+	}
+
+	out.NodeId = child.nodeID
+	out.Generation = b.nodes[child.nodeID].generation
+
+	if b.options.AttrTimeout != nil {
+		out.SetAttrTimeout(*b.options.AttrTimeout)
+	}
+	if b.options.EntryTimeout != nil {
+		out.SetEntryTimeout(*b.options.EntryTimeout)
+	}
+
+	return fuse.OK
+}
+
+func (b *rawBridge) registerInode(child *Inode) {
+	if l := len(b.free); l > 0 {
+		last := b.free[l-1]
+		b.free = b.free[:l-1]
+
+		child.nodeID = last
+		b.nodes[last].inode = child
+		b.nodes[last].generation++
+	} else {
+		last := len(b.nodes)
+		b.nodes = append(b.nodes, mapEntry{
+			inode: child,
+		})
+		child.nodeID = uint64(last)
+	}
+}
+
+func (b *rawBridge) Create(input *fuse.CreateIn, name string, out *fuse.CreateOut) (code fuse.Status) {
+	ctx := context.TODO()
+	parent, _ := b.inode(input.NodeId, 0)
+	child, f, flags, code := parent.node.Create(ctx, name, input.Flags, input.Mode)
+	if !code.Ok() {
+		if b.options.NegativeTimeout != nil {
+			out.SetEntryTimeout(*b.options.NegativeTimeout)
+		}
+		return code
+	}
+
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	b.registerInode(child)
+
+	lockNodes(parent, child)
+	parent.setEntry(name, child)
+	unlockNodes(parent, child)
+
+	out.NodeId = child.nodeID
+	out.Generation = b.nodes[child.nodeID].generation
+
+	if b.options.AttrTimeout != nil {
+		out.SetAttrTimeout(*b.options.AttrTimeout)
+	}
+	if b.options.EntryTimeout != nil {
+		out.SetEntryTimeout(*b.options.EntryTimeout)
+	}
+
+	out.Fh = b.registerFile(f)
+	out.OpenFlags = flags
+
+	f.GetAttr(ctx, &out.Attr)
+	return fuse.OK
+}
+
+func (b *rawBridge) Forget(nodeid, nlookup uint64) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	n := b.nodes[nodeid].inode
+	n.lookupCount -= nlookup
+	if n.lookupCount == 0 {
+		n.clearChildren()
+		n.clearParents()
+
+		b.free = append(b.free, nodeid)
+		b.nodes[nodeid].inode = nil
+	}
+
+}
+
+func (b *rawBridge) SetDebug(debug bool) {}
+
+func (b *rawBridge) GetAttr(input *fuse.GetAttrIn, out *fuse.AttrOut) (code fuse.Status) {
+	n, fEntry := b.inode(input.NodeId, input.Fh())
+	f := fEntry.file
+
+	// nosubmit - FATTR_FH vs FUSE_GETATTR_FH ?
+	if input.Flags()&fuse.FUSE_GETATTR_FH == 0 {
+		f = nil
+	}
+
+	dest := &out.Attr
+	code = n.node.GetAttr(context.TODO(), f, dest)
+	if out.Nlink == 0 {
+		// With Nlink == 0, newer kernels will refuse link
+		// operations.
+		out.Nlink = 1
+	}
+
+	// NOSUBMIT attr timeout
+	return code
+}
+
+func (b *rawBridge) SetAttr(input *fuse.SetAttrIn, out *fuse.AttrOut) (code fuse.Status) {
+
+	ctx := context.TODO()
+
+	n, fEntry := b.inode(input.NodeId, input.Fh)
+	f := fEntry.file
+	if input.Valid&fuse.FATTR_FH == 0 {
+		f = nil
+	}
+
+	if input.Valid&fuse.FATTR_MODE != 0 {
+		permissions := uint32(07777) & input.Mode
+		code = n.node.Chmod(ctx, f, permissions)
+	}
+
+	if code.Ok() && (input.Valid&(fuse.FATTR_UID|fuse.FATTR_GID) != 0) {
+		var uid uint32 = ^uint32(0) // means "do not change" in chown(2)
+		var gid uint32 = ^uint32(0)
+		if input.Valid&fuse.FATTR_UID != 0 {
+			uid = input.Uid
+		}
+		if input.Valid&fuse.FATTR_GID != 0 {
+			gid = input.Gid
+		}
+		code = n.node.Chown(ctx, f, uid, gid)
+	}
+
+	if code.Ok() && input.Valid&fuse.FATTR_SIZE != 0 {
+		code = n.node.Truncate(ctx, f, input.Size)
+	}
+
+	if code.Ok() && (input.Valid&(fuse.FATTR_ATIME|fuse.FATTR_MTIME|fuse.FATTR_ATIME_NOW|fuse.FATTR_MTIME_NOW) != 0) {
+		now := time.Now()
+		var atime *time.Time
+		var mtime *time.Time
+
+		if input.Valid&fuse.FATTR_ATIME != 0 {
+			if input.Valid&fuse.FATTR_ATIME_NOW != 0 {
+				atime = &now
+			} else {
+				t := time.Unix(int64(input.Atime), int64(input.Atimensec))
+				atime = &t
+			}
+		}
+
+		if input.Valid&fuse.FATTR_MTIME != 0 {
+			if input.Valid&fuse.FATTR_MTIME_NOW != 0 {
+				mtime = &now
+			} else {
+				t := time.Unix(int64(input.Mtime), int64(input.Mtimensec))
+				mtime = &t
+			}
+		}
+
+		code = n.node.Utimens(ctx, f, atime, mtime)
+	}
+
+	if !code.Ok() {
+		return code
+	}
+
+	// Must call GetAttr(); the filesystem may override some of
+	// the changes we effect here.
+	attr := &out.Attr
+	code = n.node.GetAttr(ctx, f, attr)
+
+	// TODO - attr timout?
+	return code
+}
+
+func (b *rawBridge) Mknod(input *fuse.MknodIn, name string, out *fuse.EntryOut) (code fuse.Status) {
+	return fuse.ENOSYS
+}
+
+func (b *rawBridge) Mkdir(input *fuse.MkdirIn, name string, out *fuse.EntryOut) (code fuse.Status) {
+	return fuse.ENOSYS
+}
+
+func (b *rawBridge) Unlink(header *fuse.InHeader, name string) (code fuse.Status) {
+	return fuse.ENOSYS
+}
+
+func (b *rawBridge) Rmdir(header *fuse.InHeader, name string) (code fuse.Status) {
+	return fuse.ENOSYS
+}
+
+func (b *rawBridge) Rename(input *fuse.RenameIn, oldName string, newName string) (code fuse.Status) {
+	return fuse.ENOSYS
+}
+
+func (b *rawBridge) Link(input *fuse.LinkIn, filename string, out *fuse.EntryOut) (code fuse.Status) {
+	return fuse.ENOSYS
+}
+
+func (b *rawBridge) Symlink(header *fuse.InHeader, pointedTo string, linkName string, out *fuse.EntryOut) (code fuse.Status) {
+	return fuse.ENOSYS
+}
+
+func (b *rawBridge) Readlink(header *fuse.InHeader) (out []byte, code fuse.Status) {
+	return nil, fuse.ENOSYS
+}
+
+func (b *rawBridge) Access(input *fuse.AccessIn) (code fuse.Status) {
+	return fuse.ENOSYS
+}
+
+// Extended attributes.
+func (b *rawBridge) GetXAttrSize(header *fuse.InHeader, attr string) (sz int, code fuse.Status) {
+	return 0, fuse.ENOSYS
+}
+
+func (b *rawBridge) GetXAttrData(header *fuse.InHeader, attr string) (data []byte, code fuse.Status) {
+	return nil, fuse.ENOSYS
+}
+
+func (b *rawBridge) ListXAttr(header *fuse.InHeader) (attributes []byte, code fuse.Status) {
+	return nil, fuse.ENOSYS
+}
+
+func (b *rawBridge) SetXAttr(input *fuse.SetXAttrIn, attr string, data []byte) fuse.Status {
+	return fuse.ENOSYS
+}
+
+func (b *rawBridge) RemoveXAttr(header *fuse.InHeader, attr string) (code fuse.Status) {
+	return
+}
+
+func (b *rawBridge) Open(input *fuse.OpenIn, out *fuse.OpenOut) (status fuse.Status) {
+	n, _ := b.inode(input.NodeId, 0)
+	// NOSUBMIT: what about the mode argument?
+	f, flags, code := n.node.Open(context.TODO(), input.Flags)
+	if !code.Ok() {
+		return code
+	}
+
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	out.Fh = b.registerFile(f)
+	out.OpenFlags = flags
+	return fuse.OK
+}
+
+func (b *rawBridge) registerFile(f File) uint64 {
+	var fh uint64
+	if len(b.freeFiles) > 0 {
+		last := uint64(len(b.freeFiles) - 1)
+		fh = b.freeFiles[last]
+		b.freeFiles = b.freeFiles[:last]
+	} else {
+		fh = uint64(len(b.files))
+		b.files = append(b.files, fileEntry{})
+	}
+
+	b.files[fh].file = f
+	return fh
+}
+
+func (b *rawBridge) Read(input *fuse.ReadIn, buf []byte) (fuse.ReadResult, fuse.Status) {
+	n, f := b.inode(input.NodeId, input.Fh)
+	return n.node.Read(context.TODO(), f.file, buf, int64(input.Offset))
+}
+
+func (b *rawBridge) GetLk(input *fuse.LkIn, out *fuse.LkOut) (code fuse.Status) {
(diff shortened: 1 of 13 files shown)


rfjakob:
Couple superficial comments:

This may be an opportunity to choose clearer names than `Node` and `Inode`. I always found it hard to remember which is which, and it does not help that libfuse has the naming the other way round. Our `Inode` = libfuse [`struct node`](https://github.com/libfuse/libfuse/blob/63d53ecc3addb6fcf3d7f38f2de6822ba37bfd2d/lib/fuse.c#L142).

Ideas for Inode: KernelNode? VfsNode? TreeNode?
Ideas for Node: FsNode? UserNode?

navytux:
@hanwen, thanks for trying this. Regarding locks I would suggest to try:

1. have one global lock that protects nodeID tab;
2. each inode to also have its own lock.

Then lookup is:

```
lookup(parentID, name):
    nodeidTab.lock()
    parent = nodeidTab[parentID]
    nodeidTab.unlock()

    child = parent.node.Lookup(name)
    lockNodes(parent, child)

    child.nlookup++
    parent.children[name] = child
    child.parents[(parent,name)] = parent

    unlockNodes(parent, child)
```

where `lockNodes(nodev ...Inode)` locks nodes in their address-ascending order (to prevent deadlocks) avoiding duplicates.

On `forget` we would have to lock node, extract children and parents, `unlock node`, `relock(node, children, parents)`, verify that children/parents had not change (and retry if they changed), and perform unlinking.

if we do it with only one big filesystem lock, filesystems will become likely not scalable.

----

Another observation since we are talking about API cleanup: we could require Node to provide `setInode` (unexported) with practical way to achieve this via embeding `nodefs.DefaultNode`, not `nodefs.Node` with *DefaultNode providing `setInode` (unexported) and `Inode` (exported). This way we remove SetInode from public API, make Node instances smaller (instead of 2 words for Node interface, only 1 word for *Inode pointer) with also slightly efficient inherited methods, and less pointers (thus less pressure on Go GC) as there will be no inermediate Node -> DefaultNode -> Inode. The flexibility of using other default nodes will be still there, as users will still be able to create their default types like this:

```go
type MyDefaltNode struct {
    nodefs.DefaultNode
}

func (d *MyDefaultNode) OverridenOp() {
    ....
}
```

--------

Also since  we are talking about API, and you recently mentioned jacobsa/fuse, I would suggest to consider the following:

- go-fuse strong points: FUSE protocol coverage, efficiency (buffer reuse, splice, multiple /dev/fuse workers, ...).
- go-fuse weak points: API is a bit messy, both external and internal organization.

On the other hand user-visible API of rsc/fuse family (rsc/fuse, bazil/fuse, jacobsa/fuse) is good.
I would suggest to try not to diverge from user-visible API there unless it makes practical value to diverge.

From this point of view I would suggest to split go-fuse into the following packages:

- `fuse/fusenet` (provides `Conn` to send/recive FUSE messages over /dev/fuse, and FUSE messages definitions). This is useful to implement both protocol translators and as a lower-level support to implement node-level filesystem servers. Kill `RawFileSystem`

- `fuse` this provides node-level API to implement filesystem servers (Node, Handle) and a way to mount  root Node + serve. This should be based on what nodefs is today. I suggest not to rename Node -> XXX Node and try to hide Inode from public API as much as we can.

--------

I also suggest that we release current state as v1 - not to break backward compatibility - and start to make changes incrementally, e.g. kill submounts, add context, ...

rfjakob:
I think this locking scheme has the same problem as https://github.com/hanwen/go-fuse/issues/168#issuecomment-459740379 , doesn't it?

navytux:
(@rfjakob, thanks for feedback. I'm currently overbusy. I will try to provide my reply in several days or a week; sorry)

hanwen:
let's discuss the naming and package organization when we're sure how it's going to work. It's the least important part of the dicsussion

I took a closer look at jacobsa fuse package, but it is actually built on top of rsc/bazil fuse (it embeds a copy of bazil fuse), so we're back to the rsc/bazil API.

hanwen:
regarding locking: how does libfuse do it?

I need to think about your locking scheme to see if I understand it.

I think we can't hide Inode from the callers of the library. Many of the filesystems that I wrote before have to traverse the tree, so they need access to the Inode.

hanwen:
re: versioning:

we could leave the current structure in place (with nodefs/pathfs libraries), and add a lib in a new location for the new code. This avoids forcing updates onto callers.

hanwen:
libfuse has an easier time because it keeps track of parent pointers, but not child pointers.

Is that a sensible API ? I think many FS-es will want to keep track of children separately.

Does the kernel VFS also only have parent pointers?

rfjakob:
About the kernel VFS, it does not have all the answers, but that's what I summarized some time ago: https://nuetzlich.net/the-fuse-wire-protocol/

rfjakob:
Looks like the kernel VFS has both child and parent pointers:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/dcache.h?h=v4.13-rc7#n89 :
```C
struct dentry {
    ...
    struct dentry *d_parent;	/* parent directory */
    ...
    struct list_head d_subdirs;	/* our children */
}

**Comment by hanwen** [comment]
I followed navytux' recommendation. The lookup/forget is still done under lock, so #168 has no chance of returning.

**Comment by hanwen** [comment]
PTAL - added files.

question: if we use fuse types in function signatures, how do we handle SetAttr() ?

**Comment by rfjakob** [comment]
I'm unclear what FindChildByOpaqueID is about. Shouldn't the search be filesystem-wide?

**Comment by rfjakob** [comment]
If we want to properly support hard links, we have to do that, otherwise we end up with a variant of this bug: https://github.com/vgough/encfs/issues/452

**Comment by rfjakob** [comment]
Also, having an inode number table would allow us to drop the children list from Inode (at least for filesystems that have inode numbers, like loopback or gocryptfs)

**Comment by hanwen** [comment]
I don't understand the last remark. How does that work?

**Comment by rfjakob** [comment]
On LOOKUP, we have to determine if the node is already known. If it is, we can do `nlookup++` and avoid creating a new one.

With the old API, this effectively happens twice: https://github.com/hanwen/go-fuse/blob/33711add40f027a04cd48dbfe19e5dbe5f7855fe/fuse/pathfs/pathfs.go#L543
Once via `n.Inode().GetChild(name)`, and once via `n.pathFs.clientInodeMap[fi.Ino]`. The first one, `n.Inode().GetChild(name)`, is superflous when we have stable inode numbers. And this means we can drop the children list.

**Comment by hanwen** [comment]
The user can also keep track of inode numbers.

The children list is there so you can construct FS trees in memory, without the construction having to be driven by lookup calls. Perhaps that is a bad idea, though, but it allows the default FS to provide a sensible Readdir out of the box, for example.

**Comment by hanwen** [comment]
we could do the opt-out as a later optimization maybe? I'd rather not make it more complicated than necessary.

**Comment by hanwen** [comment]
I thought a bit more about the opaque ID, and I think we can drop it. It's not that useful anyway because in loopback the uint64 isn't enough (you have to add the device ID for it to become unique.)

OTOH, maybe we should try to use it so people can use the uint64 ID for NFS exports. What do you think? 

Regardless, I added a Forgotten method (see latest version of the code) so callers have a way to do cleanup in the background.

**Comment by rfjakob** [comment]
> OTOH, maybe we should try to use it so people can use the uint64 ID for NFS exports.

You mean for a userspace NFS client? Looks like NFS file handles can be up to 64 **bytes**, so this wouldn't be enough...

**Comment by hanwen** [comment]
IIRC, an NFS server have inode numbers that persist across server reboots. You could do that by wiring through the uint64 inode numbers (and hope that those are persistent), but honestly I'm willing to dismiss NFS.

**Comment by navytux** [comment]
@rfjakob, @hanwen, first of all I appologize for the delay with replying.

I've made some review of the new nodefs here:

https://lab.nexedi.com/kirr/go-fuse/commit/7e1f11bce0?expand_all_diffs=1

I copy the summry to this message for convenience:

---- 8<----

- add nodefs package overview; in particular describe what inode is so
  that there is no confusion about its meaning.

- Draftly implement Lookup / Forget. They work not under global
  rawBridge.mu and still there should be no race of Lookup / Forget due to
  careful locking of inode -> (inode, ichild) + retrying.
  Add description of Forget semantic and what happens when we receive
  forget for a directory for which children are not forgotten yet.

  ( it is too late here now and I did not checked the implementation with
    a fresh head. I thought that it is better to release current state
    for discussion as I likely won't be able to work on newapi for at
    least another week )

- use atomics in DefaultNode setInode/inode; see Lookup and
  corresponding description nearby DefaultNode.setInode for why it is
  needed.

- inode.{lookupCount,nodeID} are now protected not by global
  rawBridge.mu, but instead by inode.mu .

- change Node operation to return Nodes, not Inode. In particulare
  Node.Lookup should now return Node. Inodes are internal index of
  nodefs VFS (see package description) and we should not load filesystem
  implementations to think about them where we can. Also it makes a
  more closed interface when filesystem works in terms it nodes
  completely. Also this way we offload filesystems for caring about
  tricky details of how to create inode for a hardlinked entry (see
  Lookup for details which handles about it)

- Remove Node.Inode -> nodefs.InodeOf(Node). this way there will be no
  possibility to override Node.Inode and we can specify InodeOf semantic
  exactly in API docs.

- unlockNodes: sort is not needed

- lock/unlock Nodes: avoid duplicates (e.g. there can be duplicates if dir/a and
  dir/b are hardlinks to the same file. If we don't avoid duplicates lockNodes will deadlock)

- made some other edits, part of them not complete...

---- 8< ----

Please have a look and consider integrating it.

P.S.

@rfjakob - there should be no race of between LOOKUP and FORGET - please see the changes for overview of how it works and why there is no race.

About kernel's `struct dentry` and .parent there - if I understand correctly -  it was for dentry, not for inode. for dir1/file and dir2/file, even if file is the same and hardlinked it will be _two_ different dentries.

P.P.S. I, once again, appologize for the delay with replying and a bit clumsy feedback.

**Comment by hanwen** [comment]
no worry about the delay. I'm happy for your in-depth comments and improvements. There are some questions I had about FORGET semantics that I left on your gitlab commit (I always see parents being forgotten with their children, does the kernel allow otherwise?)

**Comment by navytux** [comment]
@hanwen, thanks for feedback and kind words. I understand the question about FORGET semantics is correct to ask. Please se my reply about it here: https://lab.nexedi.com/kirr/go-fuse/commit/7e1f11bce0#note_74589 and in other comments on that commit.

Copying summary here for convenience:

---- 8< ----
What I did:

- I studied fs/fuse/ in Linux to see when it sends forget. What I saw was essentially that a forget is queued whenever pagecache wants to evict a node:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/fuse/inode.c?id=cb5b020a8d38#n134
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/fuse/inode.c?id=cb5b020a8d38#n815

- I did not studied how Linux evicts inodes from pagecache. But offhand to me it seems, at least theoreticlly, possible that after a file inode was looked up, opened and being used, under memory pressure its directory inode could be evicted. I admit that kernel implementation might want to keep the directory inode, but please see below

- Even though in practice FUSE is not specified and strongly tied to Linux VFS semantics, it is still more correct to think about FUSE as about network protocol in the first place. From this point of view we (filesystem server) should not blindly trust our client (usually kernel) to not misbehave and to be bug free. This is especially true since there are several FUSE clients out there (Linux, MacOS, FreeBSD, ...) and that event just for Linux for years they are having trouble to document what VFS semantics are: https://lwn.net/Articles/779747.

- Even for libfuse there is a history when kernel was sending something that was causing abort in the library, see e.g. https://git.kernel.org/linus/50322fe7d4, ...

With all those arguments, should we trust the kernel (our client) not to misbehave? Is it impossible on the protocol level to send FORGET for idir without sending FORGET for ifile that was under it?

To me the answer is: we should not blindly trust client and we should not misbehave/crash if client sends something unexpected.
That's why in my implementation FORGET(idir) is delayed untill all its children are gone.

---- 8< ----

**Comment by rfjakob** [comment]
To the question of whether nodes that have children can be forgotten, I did this experiment. Kernel = 4.20.6.
```diff
--- a/fuse/nodefs/fsconnector.go
+++ b/fuse/nodefs/fsconnector.go
@@ -150,6 +150,7 @@ func (c *FileSystemConnector) forgetUpdate(nodeID uint64, forgetCount int) {
                        node == c.rootNode || node.mountPoint != nil {
                        // We cannot forget a directory that still has children as these
                        // would become unreachable.
+                       panic("node has children")
                        return
                }
                // We have to remove ourself from all parents.
```

I was able to quickly trigger the panic with fsstress:
```
gocryptfs/tests/stress_tests$ ./fsstress-gocryptfs.bash 
Recompile gocryptfs
gocryptfs v1.7-rc2-1-gfa81c13-dirty; go-fuse v20170619-106-g58dcd77-dirty; 2019-02-16 go1.11.4
Waiting for mount:  ok
Starting fsstress loop
1
    fsstress.1 seed = 1550092374
    fsstress.2 seed = 1550731535
    fsstress.3 seed = 1550629286
    rm
2
    fsstress.1 seed = 1550060910
    fsstress.2 seed = 1550703298
panic: node has children

goroutine 67952 [running]:
github.com/hanwen/go-fuse/fuse/nodefs.(*FileSystemConnector).forgetUpdate(0xc00007acc0, 0x627, 0x2)
	github.com/hanwen/go-fuse/fuse/nodefs/fsconnector.go:153 +0x3f8
github.com/hanwen/go-fuse/fuse/nodefs.(*rawBridge).Forget(0xc00007acc0, 0x627, 0x2)
	github.com/hanwen/go-fuse/fuse/nodefs/fsops.go:133 +0x82
github.com/hanwen/go-fuse/fuse.doBatchForget(0xc0000a8000, 0xc000442000)
	github.com/hanwen/go-fuse/fuse/opcode.go:322 +0xff
github.com/hanwen/go-fuse/fuse.(*Server).handleRequest(0xc0000a8000, 0xc000442000, 0xc000442000)
	github.com/hanwen/go-fuse/fuse/server.go:431 +0x26b
github.com/hanwen/go-fuse/fuse.(*Server).loop(0xc0000a8000, 0xc0001d5a01)
	github.com/hanwen/go-fuse/fuse/server.go:403 +0x18f
created by github.com/hanwen/go-fuse/fuse.(*Server).readRequest
	github.com/hanwen/go-fuse/fuse/server.go:291 +0x2d8
fsstress: check_cwd failure
fsstress: check_cwd failure
mkdir: cannot create directory ‘/tmp/fsstress-gocryptfs.bash.YgM.mnt/fsstress.3’: Transport endpoint is not connected
```

navytux:
@rfjakob, thanks for this test. Now we know the issue is not purely theoretical.

hanwen:
this proves that there can be a (re)ordering of request handling where the parent is processed before the child. My question is wether it is possible for a child nodes to be used in requests other than LOOKUP while their parents have been forgotten.

Regardless, if we take the stance that the kernel might be wrong and forget parents before forgetting children, I need to rethink how to deal with persistent nodes. Maybe there is no way for us to let the kernel control lifetime of a (persistent) subtree by looking up and forgetting the root of the subtree.

rfjakob:
> whether it is possible for a child node to be used in requests other than FORGET while their parents have been forgotten.

Good question, I'll try to come up with an experiment.

rfjakob:
Experiment is:
```diff
diff --git a/fuse/nodefs/fsconnector.go b/fuse/nodefs/fsconnector.go
index 7299048..86e1bad 100644
--- a/fuse/nodefs/fsconnector.go
+++ b/fuse/nodefs/fsconnector.go
@@ -146,7 +146,7 @@ func (c *FileSystemConnector) forgetUpdate(nodeID uint64, forgetCount int) {
        defer node.mount.treeLock.Unlock()
 
        if forgotten, _ := c.inodeMap.Forget(nodeID, forgetCount); forgotten {
-               if len(node.children) > 0 || !node.Node().Deletable() ||
+               if !node.Node().Deletable() ||
                        node == c.rootNode || node.mountPoint != nil {
                        // We cannot forget a directory that still has children as these
                        // would become unreachable.
diff --git a/fuse/pathfs/pathfs.go b/fuse/pathfs/pathfs.go
index 96306e1..96c0c93 100644
--- a/fuse/pathfs/pathfs.go
+++ b/fuse/pathfs/pathfs.go
@@ -309,7 +309,7 @@ func (n *pathInode) GetPath() string {
                // the tree using unlink, but we are forced to run
                // some file system operation, because the file is
                // still opened.
-
+               panic("trying to use orphaned node")
                return ".deleted." + n.inode.String()
        }
```

And it triggered within 2 minutes of fsstress:

```
gocryptfs/tests/stress_tests$ ./fsstress-gocryptfs.bash 
Recompile gocryptfs
gocryptfs v1.7-rc2-1-gfa81c13-dirty; go-fuse v20170619-106-g58dcd77-dirty; 2019-02-17 go1.11.4
Waiting for mount:  ok
Starting fsstress loop
1
    fsstress.1 seed = 1550441880
    fsstress.2 seed = 1550001692
    fsstress.3 seed = 1550170782
    rm
2
    fsstress.1 seed = 1549888475
    fsstress.2 seed = 1550455867
    fsstress.3 seed = 1550370444
    rm
3
    fsstress.1 seed = 1550805877
    fsstress.2 seed = 1550604747
panic: trying to use orphaned node

goroutine 142296 [running]:
github.com/hanwen/go-fuse/fuse/pathfs.(*pathInode).GetPath(0xc000087e30, 0xc0005cc5d0, 0x5ac506)
	github.com/hanwen/go-fuse/fuse/pathfs/pathfs.go:312 +0x555
github.com/hanwen/go-fuse/fuse/pathfs.(*pathInode).GetXAttr(0xc000087e30, 0xc000427140, 0x13, 0xc001421a58, 0x8, 0x18, 0xc000bb5b00, 0x465302)
	github.com/hanwen/go-fuse/fuse/pathfs/pathfs.go:390 +0x2f
github.com/hanwen/go-fuse/fuse/nodefs.(*rawBridge).GetXAttrSize(0xc000086cc0, 0xc001421a40, 0xc000427140, 0x13, 0x44, 0x0)
	github.com/hanwen/go-fuse/fuse/nodefs/fsops.go:380 +0x72
github.com/hanwen/go-fuse/fuse.doGetXAttr(0xc0000be000, 0xc0014218c0)
	github.com/hanwen/go-fuse/fuse/opcode.go:247 +0x103
github.com/hanwen/go-fuse/fuse.(*Server).handleRequest(0xc0000be000, 0xc0014218c0, 0xc0014218c0)
	github.com/hanwen/go-fuse/fuse/server.go:431 +0x26b
github.com/hanwen/go-fuse/fuse.(*Server).loop(0xc0000be000, 0xc000bb5a01)
	github.com/hanwen/go-fuse/fuse/server.go:403 +0x18f
created by github.com/hanwen/go-fuse/fuse.(*Server).readRequest
	github.com/hanwen/go-fuse/fuse/server.go:291 +0x2d8
fsstress: check_cwd failure
fsstress: check_cwd failure
fsstress: check_cwd failure
```

hanwen:
I spent some time looking through FUSE kernel and libfuse. 

libfuse has a refctr field, detailing the number of references (the kernel counts as one) that each node has,

 https://github.com/libfuse/libfuse/blob/276728f70fc8923c90f1d2efaae8c7c8ce205b9b/lib/fuse.c#L147

so it holds on to nodes until the last reference (child) has disappeared.

I looked more at the VFS docs: I was confused because there is a dcache who entries (dentries)  keep track of children, and points to an inode. The LOOKUP operation populates a dentry + an inode, but FORGET only drops the inode. I suppose the dentries are recycled more often.

rfjakob:
> libfuse has a refctr field

I wonder if we can let the garbage collector handle this. @navytux I see you handle delayed forget explicitly, and I also wondered if we can push that work to the garbage collector, maybe with runtime.SetFinalizer if needed.

rfjakob:
No, this won't work, because the node has to stay in the children list of its parents. Otherwise we cannot keep track of it on RENAME. That's also the reason why leaving the children list nil ( https://github.com/hanwen/go-fuse/pull/265#issuecomment-462261516) can't work.

libfuse also has to find the children nodes on RENAME: https://github.com/libfuse/libfuse/blob/276728f70fc8923c90f1d2efaae8c7c8ce205b9b/lib/fuse.c#L1475
It just does so using `f->name_table` instead of having a children list in each node.

hanwen:
Kirill, I integrated some of your changes into my new push. I left bridge.mu to protect the lookup/nodeID until we have something that doesn't lock while calling user code. 

I also retained the API to use *Inode as return value. It is better to have the user code setup the Inode link, because more complex file systems may want to populate the subtree in one go, and you need to have the Inode to be able to do that. But, it should now be immune to the race you observed.

I also want to try to implement something like libfuse's refctr (but it seems hairy given how we do locking.)

navytux:
@hanwen, @rfjakob, thanks for feedback. I agree reworked Lookup is not fully ready because of inode.node.Lookup call under lock and because of Create stub. I left comment about calling Lookup not under mu here: https://lab.nexedi.com/kirr/go-fuse/commit/7e1f11bc?expand_all_diffs=1#note_74596. I need to put aside work on newapi for some time again and could hopefully get back to this in a week or two, or maybe later... I appologize for the delays here.

I will  try to look into  `because more complex file systems may want to populate the subtree in one go` requrement and see how we should maybe could do it with node operations still working in terms of Node, and still one can costruct the tree beforehand. Please post a filesystem example where such population is performed beforehand, so that the use case is more clear to me (I will try to look into this in the next batch of newapi work on my side).

It is tempting to use finalizers, aka weakrefs, but finalizers have lots of restrictions, including one that says that they will not be run if the object is part of a cycle. I suggest we do not depend on finalizers and do everything explicitly (and this comes from a person who did weakref go package and uses it for real where it is absolutely needed: https://lab.nexedi.com/kirr/neo/blob/t/go/zodb/internal/weak/weak.go https://lab.nexedi.com/kirr/neo/blob/7fbd77cb/go/zodb/connection.go#L68)

I also spent some time checking how kernel handles eviction. Here is what I used to observe how evict is called there:

---- 8< ---- (ievicttrace)
```c
#!/usr/bin/env bpftrace
# ievicttrace - trace from where / how inodes are evicted by Linux kernel

#include <linux/fs.h>

/*
kprobe:evict {
        @ievict[kstack] = count();
}

interval:s:3 {
        print(@ievict);
        clear(@ievict);
}
*/

kprobe:fuse_evict_inode {
        $inode = (inode *)arg0;
        $t = elapsed / 1000;
        printf("%d.%d EVICT i%d: %s\n", $t/1000000, $t%1000000, $inode->i_ino, kstack);
}
```
---- 8< ----

and the output looks like

```
5.901699 EVICT i5: 
        fuse_evict_inode+1
        evict+210
        __dentry_kill+213
        dentry_kill+77
        dput.part.32+203
        __fput+231
        task_work_run+138
        exit_to_usermode_loop+235
        do_syscall_64+253
        entry_SYSCALL_64_after_hwframe+68

5.901769 EVICT i4: 
        fuse_evict_inode+1
        evict+210
        __dentry_kill+213
        dentry_kill+77
        dput.part.32+203
        __fput+231
        task_work_run+138
        exit_to_usermode_loop+235
        do_syscall_64+253
        entry_SYSCALL_64_after_hwframe+68

10.935121 EVICT i3: 
        fuse_evict_inode+1
        evict+210
        __dentry_kill+213
        shrink_dentry_list+148
        shrink_dcache_parent+136
        do_one_tree+14
        shrink_dcache_for_umount+40
        generic_shutdown_super+26
        kill_anon_super+20
        deactivate_locked_super+47
        cleanup_mnt+59
        task_work_run+138
        exit_to_usermode_loop+235
        do_syscall_64+253
        entry_SYSCALL_64_after_hwframe+68

10.935166 EVICT i-1: 
        fuse_evict_inode+1
        evict+210
        __dentry_kill+213
        shrink_dentry_list+148
        shrink_dcache_parent+136
        do_one_tree+14
        shrink_dcache_for_umount+40
        generic_shutdown_super+26
        kill_anon_super+20
        deactivate_locked_super+47
        cleanup_mnt+59
        task_work_run+138
        exit_to_usermode_loop+235
        do_syscall_64+253
        entry_SYSCALL_64_after_hwframe+68

10.935184 EVICT i1: 
        fuse_evict_inode+1
        evict+210
        __dentry_kill+213
        dentry_kill+77
        dput.part.32+203
        shrink_dcache_for_umount+40
        generic_shutdown_super+26
        kill_anon_super+20
        deactivate_locked_super+47
        cleanup_mnt+59
        task_work_run+138
        exit_to_usermode_loop+235
        do_syscall_64+253
        entry_SYSCALL_64_after_hwframe+68
```

Above i5 was for file, i4 for dir containing that file etc.  Basically Linux maintains reference counts in between inodes and dentries and it seems e.g. an opened file is keeping reference on its parent directory and most of the time things are released by fput. The kernel comment for dput "If the parent dentries were scheduled for release they too may now get deleted":

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/dcache.c?id=64c0133eb88a3b0c11c42580a520fe78b71b3932#n837

tells me that the kernel a) keeps parent dentries live while dentry for file inode is live, and b) delays evicting them, probably similar to how I delayed processing directory forgets.

However I, once again, appeal for us not to depend on the exact kernel behaviour. It is easy to make kernel bugs (see previous LWN post on refcounting in kernel, cycles there, and even main Linux VFS maintainer not being  fully able to specify what is going on there), and also there are several kernel clients we should support (Linux, FreeBSD, Mac at least). For this reason delaying processing FORGET for directorises unties us from particular kernel behaviour and make us robust. I suggest not to mimic exactly what libfuse does and to do if we were serving a client we cannot fully trust.

Once again, I hope to return to work on new nodefs at some point, but it should not be very soon. I appologize for the inconvenience,

Kirill

hanwen:
for populating trees upfront, see eg. MTP-FS.

- https://github.com/hanwen/go-mtpfs/blob/master/fs/fs.go#L366

for a directory, the lookup operation has to fetch the contents of the entire directory from the device. Once it gets those entries, it might as well put the nodes for othe sibling entries there. (This makes the default readdir work correctly)

zipfs is another example: 

 https://github.com/hanwen/go-fuse/blob/master/zipfs/memtree.go#L26
 
opening a tar.gz file is expensive, so if you do that, you might as well setup all the data necessary.

slothfs (mounts Git trees as readonly FUSE)
https://github.com/google/slothfs/blob/master/fs/gitilesfs.go#L430

These examples are read-only FS'es, and 

1) probably the pattern doesn't translate well to read-write FS's
2) the data could be stored in the Node (ie. in user code) just as well.

the original nodefs API makes these examples possible. 

However, it also requires keeping track of children in the node, which complicates locking. Maybe  I was naive about what API can work for a FUSE library, and we should drop the Children field instead.

hanwen:
There are no real guarantees on if and when finalizers get called, so I agree that we should not use them.

hanwen:
BTW, the outcome of the discussion for LookupKnownChildren with @rfjakob was that maybe we should not have 'children' in the nodes.

hanwen:
@rfjakob - so it looks like libfuse has essentially a dentry cache 

  (parent, name) -> node 

and an inode cache?

  nodeId -> node

what I don't get about libfuse is that each "struct node" only has one (parent-ID, name-str). How does that work with hardlinks?

hanwen:
note to self:  

if we have a refcounter for keeping track when parents can be cleaned up, we'll have to do complicated locking anyway, since we still have to lock the child (to change the parent pointer) and the parent (to change the refcounter). So it doesn't really simplify the problem.

rfjakob:
> How does that work with hardlinks?

Surprisingly, it doesn't. Hard links in libfuse path api are broken: https://github.com/libfuse/libfuse/issues/79

hanwen:
I added a new commit that GC of nodes similar to libfuse (retain node until it has no children anymore.) 

PTAL.

nodefs/bridge.go
```diff
@@ -111,6 +110,7 @@ func (b *rawBridge) Lookup(header *fuse.InHeader, name string, out *fuse.EntryOu
 	return fuse.OK
 }
 
+// registerInode sets an inode number in the child. Must have bridge.mu
```
rfjakob: nodeid ?

rfjakob:
I'm not sure if I get how this works. The memory location can change at any time due to Go garbage collection, right?

```go
func sortNodes(ns []*Inode) {
	sort.Slice(ns, func(i, j int) bool {
		return uintptr(unsafe.Pointer(ns[i])) < uintptr(unsafe.Pointer(ns[j]))
	})
}
```
The fine-grained locking also seems to add slice allocations to a hot path, in addition to adding complexity. I mean, maybe its worth the cost for extra scalability, but at the moment it also makes it harder for me to read and understand the logic.

hanwen:
Forget isn't really a hot path. It's not synchronized, so nobody is waiting for this to finish in time.

We could specialize the lockNodes() function for the case there are 2 nodes (which is used in Lookup.)

I'll ask the Go team about garbage collection guarantees.

rfjakob:
It is true that the FORGET latency does not matter. However, goroutines don't have priorities, so the cpu cycles we spend do matter somewhat...

Anyway, let's do it in any way you like better!

hanwen:
The Go GC does not move memory, so addresses don't change.  If it were ever to grow that capability, we could add a unique ID to each Node and use that as a sorting key.

nodefs/bridge.go
```diff
@@ -0,0 +1,496 @@
...
+	parent.setEntry(name, child)
+	b.mu.Lock()
+	if child.nodeID == 0 {
+		b.registerInode(child)
+	}
+	out.NodeId = child.nodeID
+	// NOSUBMIT - or should let FS expose Attr.Ino? This makes
+	// testing semantics hard though, because os.Lstat doesn't
+	// reflect the FUSE FS
+	out.Attr.Ino = child.nodeID
```
rfjakob: Yes please let the FS decide about the the inode number. How about

```
if out.Attr.Ino == 0 {
out.Attr.Ino = \<something\>
}
```

Also I don't think the nodeID is suitable, because it gets recycled quickly. Imagine `rsync -aH` looking for hard links. Because the nodeID is quickly recycled, it will find lots of hard linked files, and the resulting copy will be corrupted..

rfjakob: Maybe
```
out.Attr.Ino = atomic.AddUint64(b.inum, 1)
```
?

(whatever we do needs to be replicated in GetAttr)

nodefs/bridge.go
```diff
@@ -38,11 +38,56 @@ type rawBridge struct {
...
+		//	    root
+		//	    /  \
+		//	  dir1 dir2
+		//	    \  /
+		//	    file
+		//
+		// dir1.Lookup("file") and dir2.Lookup("file") are executed
+		// simultaneously.  The matching FileIDs ensure that we return the
+		// same node.
+		old := b.byFileID[id]
```
hanwen: @navytux , @rfjakob  - what do you think of this approach?

nodefs/bridge.go
```diff
@@ -0,0 +1,553 @@
...
+
+	if input.Flags()&fuse.FUSE_GETATTR_FH == 0 {
+		f = nil
+	}
+
+	code = n.node.GetAttr(context.TODO(), f, out)
+	if out.Nlink == 0 {
+		// With Nlink == 0, newer kernels will refuse link
+		// operations.
+		out.Nlink = 1
```
rfjakob: Can we get rid of this? Nlink = 0 can happen. It is checked by xfstests generic/035 and currently fails: https://github.com/hanwen/go-fuse/issues/137

nodefs/syscall_linux.go
```diff
@@ -0,0 +1,20 @@
...
+
+// futimens - futimens(3) calls utimensat(2) with "pathname" set to null and
+// "flags" set to zero
+func futimens(fd int, times *[2]syscall.Timespec) (err error) {
+	_, _, e1 := syscall.Syscall6(syscall.SYS_UTIMENSAT, uintptr(fd), 0, uintptr(unsafe.Pointer(times)), uintptr(0), 0, 0)
+	if e1 != 0 {
+		err = syscall.Errno(e1)
+	}
+	return
+}
```
rfjakob: I gocryptfs, I now do this:
```go
// FutimesNano syscall.
func FutimesNano(fd int, a *time.Time, m *time.Time) (err error) {
	ts := timesToTimespec(a, m)
	// To avoid introducing a separate syscall wrapper for futimens()
	// (as done in go-fuse, for example), we instead use the /proc/self/fd trick.
	procPath := fmt.Sprintf("/proc/self/fd/%d", fd)
	return unix.UtimesNanoAt(unix.AT_FDCWD, procPath, ts, 0)
}
```
Code: https://github.com/rfjakob/gocryptfs/blob/173218895c55f381750408d9452d118b8c574f58/internal/syscallcompat/sys_linux.go#L203

Tests: https://github.com/rfjakob/gocryptfs/blob/173218895c55f381750408d9452d118b8c574f58/tests/matrix/matrix_test.go#L675

nodefs/bridge.go
rfjakob: I like having the FileID table very much. Problem (maybe?) `node` gets thrown away. The FS may have stored data in it?

rfjakob: I thought about splitting Lookup() into two phases, but that seems to just make things more complicated.

We can just document that `NewInode` may ignore `node` if it finds an older hard link.

nodefs/syscall_linux.go
hanwen: what is the problem with a syscall wrapper?

nodefs/bridge.go
hanwen: if we generate a new inode number for each call to NewInode(), we're still "wrong". You're saying it is better to pretend files that are really hardlinked are different?

We could generate the Inode number from the FileID, but the problem is that inode must be 64 bit, while FileID is 96 bits (64 for inode, 32 for device.)

hanwen: alternatively, we could do: 

- let FileID be a u64
- use this u64 as NodeID to communicate with the kernel. 

This requires an extra translation layer from (dev,ino) <-> nodeID in the loopback FS,

rfjakob: > You're saying it is better to pretend files that are really hardlinked are different?

Yes, that's extra disk space used, the other one is data loss.

rfjakob: About making FileID = nodeID = inum (inode number in stat): Actually, sounds good. The additional (dev,ino) <-> nodeID map in loopbackfs pays for itself because bridge can drop the the fileID <-> nodeID map `byFileID`.

OTOH, the kernel and libfuse people thought it is useful that nodeID and inum can have different values. Why? Maybe it's just so that libfuse can put a raw memory pointer into nodeID, without exposing it in the stat() inum, which would be ugly for several reasons.

We don't want to use raw memory pointers as nodeID, so I guess we can drop the disctinction between nodeID and inum, and simplify the whole thing a little.

If an FS does not want to use hard links, or cannot provide stable inums, it should return zero, and bridge can do `Ino = atomic.AddUint64(b.inum, 1)`. But I wonder if the kernel gets unhappy if it always gets a new nodeID for each LOOKUP. Will test.

hanwen: there is an additional factor, which is the cost of a map[uint64] vs lookup in []inodeEntry

it's probablly fast enough to not be relevant, though.

hanwen: for loopbackFS, maybe we can do something like

  (dev << 32 | inode64)

this would work well for filesystems that hand inode numbers sequentially. 

(would not work well for stacking go-fuse FSes)

rfjakob: That's true, the nodeID lookup is super-fast at the moment due to the tight packing.

For the inode64 mapping, my thought for gocryptfs was to check the device number of the root dir, and pass all inode numbers with this device number unchanged. All other device numbers could get the (dev << 32 | inode64) treatment.

Once stacking starts, this needs a proper map, but that's not too expensive: 96-bit key -> 64-bit value = 20 bytes per inode number?

hanwen:
whoops, merged by accident. REverted the merge.

nodefs/bridge.go
hanwen: I used to have raw pointers in the NodeID field, and pass them on as Ino in Stat. There were 2 problems:

- the race detector gets really confused
- older 32-bit software (notably some JDKs) would barf with ERANGE when trying to stuff a 64 bit ino into a 32 bit stat_t.

nodefs/bridge.go
hanwen: fixed now.

nodefs/bridge.go
rfjakob: EncFS also used to use raw pointers [till recently](https://github.com/vgough/encfs/commit/e2f0f8e3c68604a46d8665c3b0125bb35a0d9181). It also had a subtle bug with concurrent file open, which meant a node that still was in use by the kernel could get deallocated. Was fun to debug because it caused random segfaults: https://github.com/vgough/encfs/issues/214 . 0 of 5 stars for raw pointers from me.

nodefs/syscall_linux.go
rfjakob: No specific problem. My solution is not even less ugly, ignore the comment.

PR #269, opened 2019-02-16
fuse: Enable parallel lookup and readdir by default

navytux:
After Linux VFS was enhanced to be able to do several lookups and
readdirs in parallel, FUSE client in Linux, in fear to break existing
filesystems, explicitly reverted it back to be serial in FUSE exchange
and requested filesystems to indicate via CAP_PARALLEL_DIROPS flag that
the filesystem server is ready to handle parallel lookup and readdir
requests.

	https://git.kernel.org/linus/5c672ab3f0

We should be ready to handle and we provide MountOptions.SingleThreaded
for those filesystems, which are not ready to handle requests in
parallel.


navytux:
(CI: go1.11 failed in TestUnionFsCopyChmod)

navytux:
Thanks.

PR #270, opened 2019-02-24
Nodefs api

hanwen:
I pushed to the wrong branch by accident, which merged the open PR.

This one continues the previous PR.


rfjakob:
Yes I am ok with that.

hanwen:
@navytux , @rfjakob - has anyone ever looked at interrupt handling? I added a small test, but don't get the INTERRUPT on the operation that is blocking.

rfjakob:
Just managed to get an INTERRUPT on WRITE by `kill -9`ing the blocked dd process:
```
2019/02/24 20:31:02 rx 42: WRITE i3  1234b
2019/02/24 20:31:13 rx 43: INTERRUPT i0 
2019/02/24 20:31:13 Unimplemented opcode INTERRUPT
2019/02/24 20:31:13 tx 43:     38=function not implemented
```

rfjakob:
But not for READ, same as you saw:
```
2019/02/24 20:35:29 rx 16: OPEN i3 {O_RDONLY,0x8000} 
2019/02/24 20:35:29 tx 16:     OK, {Fh 2 }
2019/02/24 20:35:29 rx 18: FLUSH i3 {Fh 2} 
2019/02/24 20:35:29 tx 18:     OK
2019/02/24 20:35:29 rx 20: READ i3 {Fh 2 [0 +16384)  L 0 RDONLY,0x8000} 
(kill -9)
2019/02/24 20:35:37 rx 22: FLUSH i3 {Fh 2} 
2019/02/24 20:35:37 tx 22:     OK
(unblock after 100 seconds)
2019/02/24 20:37:09 tx 20:     OK,  16384b data "\x00\x00\x00\x00\x00\x00\x00\x00"...
2019/02/24 20:37:09 rx 24: RELEASE i3 {Fh 2 0x8000  L0} 
2019/02/24 20:37:09 tx 24:     OK

```

rfjakob:
Question, because it was not mentioned so far: Plan is to also add a path-based API, right?

hanwen:
Re: path API: since Inode now has a GetPath() method, there is no need for a separate Path API anymore.

rfjakob:
That's true, however, pathfs also handles some other the low-level details for the user:

1) Find an open file for a nodeid (this is now in loopbackfs in `fGetAttr`)
```go
	n.mu.Lock()
	defer n.mu.Unlock()
	for f := range n.openFiles {
		if f != nil {
			return f.GetAttr(ctx, out), true
		}
	}
```
2) Creation of a new Inode if needed (MKDIR, now in loopbackfs. Same in MKNOD, LOOKUP):
```go
	node := n.rootNode.newLoopbackNode()
	ch := n.inode().NewInode(node, out.Attr.Mode, idFromStat(&st))
```
3) Tracking open files (now in loopbackfs CREATE):
```go
	node := n.rootNode.newLoopbackNode()
	ch := n.inode().NewInode(node, st.Mode, idFromStat(&st))
	lf := newLoopbackFile(f)
	n.mu.Lock()
	defer n.mu.Unlock()
	n.openFiles[lf] = struct{}{}
	return ch, lf, 0, fuse.OK
```
4) Handle RENAME, now in loopbackfs:
```go
	var newParentLoopback *loopbackNode
	if r, ok := newParent.(*loopbackRoot); ok {
		newParentLoopback = &r.loopbackNode
	} else {
		newParentLoopback = newParent.(*loopbackNode)
	}
```

This is all code that will be shared by loopbackfs and filesystems like gocryptfs, or anybody how wants to move from the libfuse "highlevel" API.

I'd argue that it makes sense to pull this into the library, and I volunteer for doing it.

navytux:
@hanwen, thanks for feedback and examples. I see the point of being able to construct a tree of nodes on just single lookup. Such cases still could be handled if we use only Node (not Inode) in the interface, e.g. via the following:


```go
func (dir *MyDir) Lookup(name string) Node {
	idir := nodefs.InodeOf(dir)
	children := d.fetch()
	for name, child := range children {
		ichild := nodefs.InodeOf(child)
		idir.AddChild(name, ichild)
	}

	return children[name]
}
```

This requires that `InodeOf(node)` create a new Inode, if inode for node was not yet created. We can handle that with atomics just fine and still keep the fast past to be only 1 regular read. By the way I use go-mtpfs from time to time to exchange files with phone and SlothFS looks interesting (for the reference I once did something similar - a filesystem to expose git as read-only checkouts and to get file timestamps there was initial `git log --reverse --raw --no-abbrev --no-renames -c` scan (https://github.com/git/git/commit/72441af7c4 https://github.com/git/git/commit/7195fbfaf5)) ; thanks again for the links.

On related topic I see that my atomic stuff from `DefaultNode.{setInode,inode}` was semantically moved into rawBridge.newInode and the way to maintain property of 1-1 Node-Inode relies there on uniqueness of id.Ino. This however works only if the original Node has !zero FileID and will break for automatic IDs.
Because of this and given that `InodeOf(node)` should be also part of 1-1 game (see above) I suggest we should move back to my atomic approach into `.setInode/.inode/inodeOf`.

I ran out of time trying to review newapi. However below are my draft notes on topics raised in this discussion before:

- inode: have .children - is it ok?

  -> it is better we have children since one of nodefs design goal is to provide tree organization, and children allows to navigate through the tree.


- not reusing autoallocated nodeID: yes - better; I actually suggest we keep nodeID if kernel forgets a node - at least until the node remains in the tree, and that we allocate new auto nodeID in always increasing order - e.g. like pid, so that there is much less probability that the same nodeID will be used for different nodes at two close time moments.

- return new nodeID on each LOOKUP: we can return new nodeID only if previously the node was forgotten. Otherwise by returning different nodeID we tell to kernel: "this is different file" and the kernel completely drops all previous caches.

- LookupKnownChildren - both behaviours are needed:
  * if a filesystem does not handle/receive invalidations from the backend - forcing relookup/regetattr/... of already known non-expired node is better.
  * if a filesystem does have proper channel to receive backend invalidations - forcing relookup/regetattr/... of already known non-expired node might be not desirable.

  I suggest we have FS-wide default flag and allow a node to overwrite it.

  By the way, I also suggest we add a way to be able to control entry and attributes timeout per returned node. Some filesystem entries are known to stay longer, or forever, and its a waste of resources if FUSE forces relookup/regetattr every 1 second on them.

- current design feels to be too focused on loopback-kind of filesystems that wrap existing OS-level fs.

I confirm that we can use addresses of inodes for comparison since Go's GC is currently non-moving. Actually Go runtime itself uses this approach in select implementation where it needs to lock channels in the same order to avoid deadlocks:

https://github.com/golang/go/blob/go1.12beta2-153-g73b803ee53/src/runtime/select.go#L159
https://github.com/golang/go/blob/go1.12beta2-153-g73b803ee53/src/runtime/select.go#L495

Some of my draft changes corresponding to the comments are here:

https://lab.nexedi.com/kirr/go-fuse/commit/bedd3b65?expand_all_diffs=1#d58b58a43771d0ed4fb9b0fc498fbb3e96a6ab2a_36_36
(na1.patch there contains not-yet integrated back changes from first newapi review)

Apologize for the delay and for incomplete reply,
Kirill

hanwen:
regarding LookupKnownChildren:

every FS could implement that by itself in a straightforward way.  

func  (n*Operations) Lookup(string) .. {
   ch := n.GetChild("name")
   if ch != nil {  return ch }
    ..
}

why do we need the library to provide the option? The option will be hard to discover make the overall lib more difficult to understand.

hanwen:
current design already allows FS-es to control individual timeouts. The Options struct has pointers to time.Duration, setting them to nil lets the Operations control the timeouts.

hanwen:
"feels to be too focused on loopback-kind FS" 

is that because of the central role of FileID in deduping things? Or is there something else?

hanwen:
Good point about the atomics: I will put them back in.

I don't think we can use InodeOf though. To create Inode, you have to know the type of the node (for the children map), and InodeOf() doesnt know. I think it is better to have InodeOf() panic if the Inode is not set.

hanwen:
Regarding making a library for PathFS: I'm not completely convinced it is necessary, but let's first build out the nodefs library completely, so we can see more clearly what is needed.

hanwen:
I wonder if it is possible to merge DefaultOperations and Inode. Let me try.

hanwen:
We would have to let the user set {Persistent,Mode,FileID,Ops} on creation, but ensure they aren't changed afterwards.

This requires constant structs, or final values for struct members which Go doesn't support.

navytux:
@hanwen, thanks for feedback.

Re LookupKnownChildren:

It is better we keep the common behaviour in the VFS and offload filesystem implementations from that.
We can have LookupKnownChildren=y to be the default and allow filesystems to opt-in into LookupKnownChildren=n (with another option name). It's like direct IO but inverted - by default filesystems receive all LOOKUP requests, but if they are fine with carefully invalidating the inode tree, they can rely on the cached tree structure. Also its not that simple as you write as the check for child being already present there is potentially racy and has to be made very carefully in race-free manner. For this reasons I suggest to have the mode for LOOKUP to first check already existing tree to be there in nodefs VFS, not being pushed onto filesystems implementer. For option name maybe LookupUseChildren could do (not entirely happy with that name yet).

> current design already allows FS-es to control individual timeouts. The Options struct has pointers to time.Duration, setting them to nil lets the Operations control the timeouts.

Good point, thanks. Maybe we could also have !nil defaults and a way for some nodes to override that. From my point of view that would be more convenient to use instead of forcing every node to care about timeouts explicitly.

> "feels to be too focused on loopback-kind FS"
>
> is that because of the central role of FileID in deduping things? Or is there something else?

It was just my feeling on, yes, that nodefs design keeps loopback-style filesystems in mind (e.g. mapping OS inode to FUSE inode, etc). Also e.g. that the case of creating several file handles is not handled right (https://lab.nexedi.com/kirr/go-fuse/commit/bedd3b6579?expand_all_diffs=1#032b161aa8ee6058475d461672a41712f7826031_426_428). Maybe something else. But I cannot prove it and have only feelings for now. However keeping in mind both use cases: loopback-style and synthetic filesystems too would be better when thinking the API over. For the reference for FileID I suggest something like this: https://lab.nexedi.com/kirr/go-fuse/commit/bedd3b6579?expand_all_diffs=1#d58b58a43771d0ed4fb9b0fc498fbb3e96a6ab2a_84_90 (not yet fully thought out).

> Good point about the atomics: I will put them back in.

thanks.

>  don't think we can use InodeOf though. To create Inode, you have to know the type of the node (for the children map), and InodeOf() doesnt know. I think it is better to have InodeOf() panic if the Inode is not set.

That would imho be hard to use and would prevent InodeOf usage in a manner to construct whole new tree on single point lookup (see beginning of my previous message). If "To create Inode, you have to know the type of the node (for the children map)" is all that is blocking - it suggests we should just add another method to Node to report its type, no? Then things would be much more straightforward, instead of being uncertain whether particular InodeOf call might panic or not.

hanwen:
I don't like methods for NodeID or Mode, because the mode and ID are constant across the lifetime of the node. By consuming them as arguments to a constructor, that is obvious.

hanwen:
>Also its not that simple as you write as the check for child being already present there is potentially racy and has to be made very carefully in race-free manner."

1. Can you given an example of such a race?

2. How would you prevent races? Even if we no know the child, we'd have to call child.GetAttr() to get attrbutes (timestamps, size), and we can't do that under lock.

hanwen:
> Also e.g. that the case of creating several file handles is not handled right

how so ?  There is a XXX which was a question, but I think it is handled correctly.

navytux:
races: I was referring to the lock/unlock lock(iparent, ichild) that my version of lookup was doing. However maybe I missed something here since that comment was put without much thought and investigation. We would need to call only node.NodeType() under lock and full node.GetAttr() call could be done with locks released. You said you dislike having .NodeType() and I understand there is some sense in your reason. Maybe we can imitate the immutability via requiring to create DefaultNode with nodeType specified, and then it becomes .nodeType() which cannot be affected by particular node after creation and thus cannot change. I do not insist on this in any way - I just dumped my draft thoughs into the review patch after quickly going through.

> how so ? There is a XXX which was a question, but I think it is handled correctly.

Sorry, by the comment I got the impression it is not handled correctly.

In general it feels to me that it is better to spend more time next try instead of trying to provide lesser quality feedback earlier...

navytux:
( another option would be to split interfaces and then if a Node additionally implements Directory (with .Lookup) - then we know it is a directory and we know we have to create children map )

hanwen:
oh, that's a nice idea, but if you look at loopback FS, it actually complicates things a little, because you'd have to implement many types rather than one big loopbackNode.

I'm also not sure how this work with device files. I don't think you can distinguish from regular files.

navytux:
Yes, there will have to be two types: loopbackDir and loopbackFile with the filesystem creating one or another depending from underlying OS stat & S_IFDIR. For devices - there is no need to distinguish from regular files as the only difference is mode bits. So we only have two kind of nodes: leaf nodes for regular files, devices, pipes, etc, and intermediate nodes for directories, which, additionally to what files provide also provide directory-level operations like LOOUP, UNLINK, RENAME, CREATE, ...

Btw, I suggest to consider splitting: Node -> Node + Directory + File (e.g. it is possible to Open a directory, but Read on that file descriptor gives EISDIR). I also suggest we rename current File to FileHandle or just Handle (because maybe it can be used over opened Directory too).

Btw2, for nodefs.Directory I suggest we rename OpenDir -> ReadDir (old nodefs-level OpenDir just returns complete []of dir entries)

hanwen:
>Btw, I suggest to consider splitting: Node -> Node + Directory + File (e.g. it is possible to Open a >directory, but Read on that file descriptor gives EISDIR). 

i know it looks nicer, but what problem does this solve?

>I also suggest we rename current File to >FileHandle or just Handle (because maybe it can be used over opened Directory too).
 
I did this.

navytux:
> i know it looks nicer, but what problem does this solve?

If we split Node into Node and Directory (to be able to detect directory nodes just by their type as discussed in https://github.com/hanwen/go-fuse/pull/270#issuecomment-468728268) putting file-specific methonds into File looks reasonable and creates more structure thus helping people to understand the library easier.

By the way on interfaces topic I suggest: there are 2 polar approaches - have one big Node interface and have dedicated interface for every method a-la as in rsc/fuse. In between those there is some middle ground with what I was talking about Node + Directory + File that does not create too much interfaces while still having a better structure and helping (with Directory) to VFS.

Btw2, I've just noticed (I was still on newapi branch, not on nodefs-api) that Node was renamed into Operations to "decrease confusion between Inode and Node". Personally I find that rename creating more confusion than it tries to solve. I would keep Node intact and if we need to distinguish more consider renaming Inode to e.g. IndexNode, TreeNode or something like that, but keep Node named as it is. I would even consider to keep current Inode and Node - with enough package-level intruduction that should not be confusing. All this, of course, is just my personal subjective opinion and taste.

> I did this.

Thanks.

navytux:
(merging Inode into Node completely would be also good - we already require to embed either DefaultFile or DefaultDirectory and those can carry on .parents, .children, etc)

rfjakob:
> I would even consider to keep current Inode and Node

I like it when the name already tell you what is what. Node and Inode did not do a good job for me...

rfjakob:
Can we hide *Inode from the nodefs API, and let bridge handle the Inode wrangling? Maybe like this? https://github.com/rfjakob/go-fuse/commit/71f93c2b07172d8007aff936d566049ded0f944a

(compiles, but fails the tests at the moment. Wanted to get early feedback)

hanwen:
> Can we hide *Inode from the nodefs API, and let bridge handle the Inode wrangling? Maybe like this? [rfjakob@71f93c2](https://github.com/rfjakob/go-fuse/commit/71f93c2b07172d8007aff936d566049ded0f944a)
> 
> (compiles, but fails the tests at the moment. Wanted to get early feedback)

how would you prepopulate trees in this scheme?

rfjakob:
Like this maybe?
```
node0 := DefaultOperations{}
NewNodeFS(node0, ...)
node1 := DefaultOperations{}
InodeOf(node0).AddChild("foo", node1, [...])
node2 := DefaultOperations{}
InodeOf(node1).AddChild("bar", node2, [...])
```

rfjakob:
(modulo the naming question. Maybe

    nodefs.AddChild(node0, "foo", node1, ...)

would be better, to avoid exposing `Inode` at all. Same for `GetPath` etc)

hanwen:
added directory reading. Appreciate any comments.

nodefs/simple_test.go
```diff
@@ -492,6 +493,39 @@ func TestNotifyEntry(t *testing.T) {
...
+// XXX Test NotifyDelete?
 
-// Test NotifyDelete?
+func TestReadDir(t *testing.T) {
+	tc := newTestCase(t)
+	defer tc.Clean()
+
+	// XXX what about ".." and "." ?
+	want := map[string]bool{}
+	for i := 0; i < 2; i++ {
```
rfjakob: Typo, should be 110 instead of 2?

However, increasing to 110 fails in tests:
```
$ go test
20:12:20.916028 killing subprocess
--- FAIL: TestReadDir (0.02s)
    simple_test.go:517: ReadDir: readdirent: input/output error
FAIL
exit status 1
FAIL	github.com/hanwen/go-fuse/nodefs	0.448s
```

nodefs/bridge.go
```diff
@@ -517,19 +525,120 @@ func (b *rawBridge) Fallocate(input *fuse.FallocateIn) (code fuse.Status) {
...
+		if f.dirStream != nil {
+			f.dirStream.Close()
+		}
+		str, code := inode.node.ReadDir(context.TODO())
+		if !code.Ok() {
+			return code
+		}
+
+		f.hasOverflow = false
+		f.dirStream = str
```
rfjakob: Does this need locking (or an atomic pointer swap?)

nodefs/bridge.go
```diff
@@ -517,19 +525,120 @@ func (b *rawBridge) Fallocate(input *fuse.FallocateIn) (code fuse.Status) {
...
 
 func (b *rawBridge) ReadDir(input *fuse.ReadIn, out *fuse.DirEntryList) fuse.Status {
-	return fuse.ENOSYS
+	n, f := b.inode(input.NodeId, input.Fh)
+
+	if code := b.getStream(input, n, f); !code.Ok() {
+		return code
+	}
+
+	if f.hasOverflow {
```
rfjakob: Does this need locking against concurrent ReadDirs?

nodefs/bridge.go
```diff
@@ -517,19 +525,120 @@ func (b *rawBridge) Fallocate(input *fuse.FallocateIn) (code fuse.Status) {
...
+		if !code.Ok() {
+			if b.options.NegativeTimeout != nil {
+				entryOut.SetEntryTimeout(*b.options.NegativeTimeout)
+			}
+		} else {
+			b.addNewChild(n, e.Name, child, nil, entryOut)
+			b.setEntryOutTimeout(entryOut)
+			if (e.Mode &^ 07777) != (child.mode &^ 07777) {
+				// XXX should go back and change the
+				// already serialized entry
```
rfjakob: Can we move `out.AddDirLookupEntry` after `n.node.Lookup`? This would avoid the problem

nodefs/bridge.go
hanwen: I don't think so, because fuse/nodefs doesnt have locking there either.

nodefs/bridge.go
hanwen: no, because if the entry doesn't fit, we'll have done a LOOKUP (with associated nlookup increment) without telling the kernel.

nodefs/simple_test.go
hanwen: there is something wrong with the encoding or size (the failing read is exactly 4096 bytes. if we stop one entry earlier it succeeds.)

hanwen: fixed it.

nodefs/bridge.go
hanwen: oops, there is locking, in fact.

But the race detector says nothing. hmm.

hanwen:
added: 
* cancellation
* xattrs
* streaming directory reads

this now feature complete (except for ACCESS).

I thought a bit more: we could do automatic association with embedded inodes like this:
```
  type DefaultOps struct {
     Inode
  }
 
   func (n *DefaultOps) inode() *Inode {
       return &n.Inode
   } 
..


  n := &myNode{
    DefaultOps{Inode{Mode: S_IFREG, Ino: 123}}
```

we could use the *Inode address as the ID to use with the kernel.

But it feels a little ugly to me: the API is not fluent (ie., you have to read the docs to understand to set Mode & Ino).

What do you think?

hanwen:
implemented ACCESS.

navytux:
@hanwen, I just came here to briefly check, so not a full review, sorry. However one note about interrupt handling: 

```diff
-func (c *rawBridge) Fsync(input *fuse.FsyncIn) fuse.Status {
+func (c *rawBridge) Fsync(cancel <-chan struct{}, input *fuse.FsyncIn) fuse.Status {
```

I suggest to consider that we start from std `context.Context` instead of raw cancel. Reason:

Even though fuse.Context implements context.Context interface, whenever request handler derives another cancellable context from that, the implementation in stdlib type asserts on parent context and if it is one of the types it knows - it will glue parent / child contexts with pointers instead of spawning goroutine for cancel propagation:

https://github.com/golang/go/blob/ef891e1c83/src/context/context.go#L245-L265

(I'm struggling about this fact and that there is no way to hook in from outside for my [xcontext.Merge](https://godoc.org/lab.nexedi.com/kirr/go123/xcontext#hdr-Merging_contexts) as well)

hanwen:
I'm worried that API contract for context.Context is not restrictive enough, and that we might have to allocate a new one for every request.

I also expect that most FS's won't do cancellation propagation, and we'd make those pay unnecessarily.

Also *fuse.Context is an existing type used in the fuse/nodefs library function signatures. The current implementation has the advantage that it's backward compatible.

navytux:
> I'm worried that API contract for context.Context is not restrictive enough

Could you please clarify?

> and that we might have to allocate a new one for every request.

We can do the same trick that we currently do: for a request derive WithCancel from background, keep that context in a struct and remember whether we cancelled it (as only fuse library can cancel it on receiving INTERRUPT). If at the end of request we see that context was not cancelled - we can reuse it for next request. For completeness: we would need to derive WithCancel from background, and from that derive WithValue to set FUSE request attributes like UID/GID.

> I also expect that most FS's won't do cancellation propagation, and we'd make those pay unnecessarily.

Some filesystem do not need this, while others do need. For example in my filesystem, when there is READ for a block for one particular storage format, the filesystem issues 512 simultaneous network loads with smaller 4K blocks (filesystem-level block is 2M). Those loads are issued under `errgroup.WithContext` to shutdown the whole operation on a first error, and `errgroup.WithContext` uses `context.WithCancel` internally.

https://lab.nexedi.com/kirr/wendelin.core/blob/6ffc9a5286/wcfs/zblk.go#L229-340

I think this is not the only example when a networked filesystem might want to derive WithCancel from ctx passed to request handler.

> Also *fuse.Context is an existing type used in the fuse/nodefs library function signatures. The current implementation has the advantage that it's backward compatible.

I understand this argument. But we already changed signature of nodefs-level operations to accept `context.Context` e.g.

```
        Lookup(ctx context.Context, name string, out *fuse.EntryOut) (*Inode, fuse.Status)
```

isn't it? Imho we can make the transition to be even backward-compatible at raw fuse level if we store the context in raw fuse request. This approach is considered to be a crime, but if backward compatibility is a must and we are indeed trying not to loose performance, from my point of view it should be ok.

hanwen:
my plan is to have fuse/nodefs and nodefs (the new one) exist next to each other. This is also why I want a better name for the new nodefs (suggestions?).  (or should I make a new package?)

regarding API contract: for reusing data, no function may hold on to a context.Context when the request handler returns. Maybe that is implicit in the phrase "request scoped", but it is not spelled out.

regarding goroutines: I took a quick look at errgroup, but for your case, there would be one extra cancellation goroutine for the errgroup, and then 512 goroutines to do the fetches, no? In that case, the extra goroutine seems like it would be a small amount of overhead.

hanwen:
storing Context in raw fuse request: this is asking for trouble. We "read" the input by reinterpreting the raw bytes with a pointer cast. If we do that for a field that is a context.Context interface pointer, we can cause the GC to follow garbage pointers.

navytux:
Yes, maybe my use case with 512/1 ratio is not a good example. However there could be other situations where ratio is something like 2/1 or 3/1 and there that additional goroutine would be felt. I was not meaning to store Context in raw fuse request - I was meaning to store it inside `fuse.request`. 

I see about API contract. If we decide to go via Context way we should document that fuse infrastructure reuses contexts and that a handler can no longer reference it after it complets. But isn't it already the case with everything else - e.g. with other pointers passed to Getattr or Lookup - for example `out *AttrOut`? We already require that those handler don't touch pointed data after handler completes as that memory might be reused for next handler. I think we should just document that clearly at fuse/nodefs level.

I understand and agree for keeping both old and new nodefs packages not to break backward compativility. I suggest we keep `nodefs` name but put it under e.g. `v2/` prefix (just under a directory). I suggest we don't fork it into another git repository as it is easier to work / grep on semantically close things together. I'm not settled with naming on my side but offhand keeping `nodefs` or maybe `treefs` name should do.

hanwen:
@navytux  - I split out the interface between file, directory, etc. It looks pretty nice.

hanwen:
I am pretty satisfied with this API. 

Last chance to review.

navytux:
@hanwen, thanks for working on this. I'm completely out of time now, so unfortunately don't have a chance to review your work. However one thing I quickly noticed was that InodeOf was dropped and instead Inode() is in public interface with comment that it must not be overwritten. This move feels to me as a mistake.

hanwen:
I think it's odd to have both ops.Inode() and InodeOf(). 

All the internal code actually uses inode(), so it's impervious to overwriting Inode().

PR #273, opened 2019-03-15
fuse: allow filesystems to have precise control over data cache

navytux:
See https://lwn.net/ml/linux-fsdevel/20190315212556.9315-1-kirr%40nexedi.com/ for rationale and details.

PR #274, opened 2019-03-17
Fix TestFopenKeepCache

navytux:
Please see details in the last patch.

### DIFF
--- a/fuse/api.go
+++ b/fuse/api.go
@@ -154,6 +154,16 @@ type MountOptions struct {
 	// If set, ask kernel to forward file locks to FUSE. If using,
 	// you must implement the GetLk/SetLk/SetLkw methods.
 	EnableLocks bool
+
+	// If set, ask kernel not to do automatic data cache invalidation.
+	// The filesystem is fully responsible for invalidating data cache.
+	//
+	// XXX for Linux ExplicitDataCacheControl currently disables data cache
+	// to be automatically invalidated only on file mtime change. If file size
+	// changes, Linux currently unconditionally discards whole cache of the
+	// file. See https://github.com/hanwen/go-fuse/pull/273 for kernel +
+	// go-fuse patches with corresponding fixes.
+	ExplicitDataCacheControl bool
 }
 
 // RawFileSystem is an interface close to the FUSE wire protocol.
--- a/fuse/nodefs/fuse.go
+++ b/fuse/nodefs/fuse.go
@@ -8,18 +8,22 @@ import (
 	"github.com/hanwen/go-fuse/fuse"
 )
 
-// MountRoot mounts a filesystem with the given root node on the given directory.
+// Mount mounts a filesystem with the given root node on the given directory.
 // Convenience wrapper around fuse.NewServer
-func MountRoot(mountpoint string, root Node, opts *Options) (*fuse.Server, *FileSystemConnector, error) {
-	conn := NewFileSystemConnector(root, opts)
-
-	mountOpts := fuse.MountOptions{}
-	if opts != nil && opts.Debug {
-		mountOpts.Debug = opts.Debug
-	}
-	s, err := fuse.NewServer(conn.RawFS(), mountpoint, &mountOpts)
+func Mount(mountpoint string, root Node, mountOptions *fuse.MountOptions, nodefsOptions *Options) (*fuse.Server, *FileSystemConnector, error) {
+	conn := NewFileSystemConnector(root, nodefsOptions)
+	s, err := fuse.NewServer(conn.RawFS(), mountpoint, mountOptions)
 	if err != nil {
 		return nil, nil, err
 	}
 	return s, conn, nil
+}
+
+// MountRoot is like Mount but uses default fuse mount options.
+func MountRoot(mountpoint string, root Node, opts *Options) (*fuse.Server, *FileSystemConnector, error) {
+	mountOpts := &fuse.MountOptions{}
+	if opts != nil && opts.Debug {
+		mountOpts.Debug = opts.Debug
+	}
+	return Mount(mountpoint, root, mountOpts, opts)
 }
--- a/fuse/opcode.go
+++ b/fuse/opcode.go
@@ -89,11 +89,25 @@ func doInit(server *Server, req *request) {
 	server.reqMu.Lock()
 	server.kernelSettings = *input
 	server.kernelSettings.Flags = input.Flags & (CAP_ASYNC_READ | CAP_BIG_WRITES | CAP_FILE_OPS |
-		CAP_AUTO_INVAL_DATA | CAP_READDIRPLUS | CAP_NO_OPEN_SUPPORT | CAP_PARALLEL_DIROPS)
+		CAP_READDIRPLUS | CAP_NO_OPEN_SUPPORT | CAP_PARALLEL_DIROPS)
 
 	if server.opts.EnableLocks {
 		server.kernelSettings.Flags |= CAP_FLOCK_LOCKS | CAP_POSIX_LOCKS
 	}
+
+
+	dataCacheMode := input.Flags & CAP_AUTO_INVAL_DATA
+	if server.opts.ExplicitDataCacheControl {
+		// XXX this only disables automatic invalidations on mtime
+		// change, but not on size change.
+		//
+		// TODO explicitly disable invalidations on size change when
+		// kernel has proper support. Details:
+		//
+		//	https://github.com/hanwen/go-fuse/pull/273
+		dataCacheMode = 0
+	}
+	server.kernelSettings.Flags |= dataCacheMode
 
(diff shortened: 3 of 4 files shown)


navytux:
(I'm wondering why we stopped doing travis CI)

fuse/api.go
```diff
@@ -154,6 +154,16 @@ type MountOptions struct {
...
+
+	// If set, ask kernel not to do automatic data cache invalidation.
+	// The filesystem is fully responsible for invalidating data cache.
+	//
+	// XXX for Linux PreciseDataCacheControl currently disables data cache
+	// to be automatically invalidated only on file mtime change. If file size
+	// changes, Linux currently unconditionally discards whole cache of the
+	// file. See https://github.com/hanwen/go-fuse/pull/273 for kernel +
+	// go-fuse patches with corresponding fixes.
+	PreciseDataCacheControl bool
```
hanwen: I'd rather think of a generic way in which filesystems can set or alter CAP_XXX flags directly.

navytux: While I probably understand your reason here, after inspecting FUSE flags in linux/fuse.h again, I still suggest we offload go-fuse users from all picky details of FUSE protocol. Many flags were added there in the fear of breaking existing filesystems, but many those flags are handled at go-fuse level and do not directly affect go-fuse users at all. For example FUSE_SPLICE_* is whether /dev/fuse supports splice or not, FUSE_BIG_WRITES is whether filesystem can handle writes > 4K - go-fuse supports it and filesystems can request max write size to be limited via MountOptions.MaxWrite, FUSE_DO_READDIRPLUS - go-fuse cares to do both ReadDir + Lookup, FUSE_NO_OPEN_SUPPORT - it is handled at go-fuse level, FUSE_ABORT_ERROR - similar - etc.

I still suggest we extend MountOption with semantic that the filesystem is fully respondible to invalidate the data cache. And then go-fuse does its best to negotiate that semantic to kernel given kernel capabilities it detects at runtime.

Filesystems that are very picky on which capabilities have been selected can already use Server.KernelSettings to retrieve and check those.

We can, of course, add a func to mount opttions that receives client capabilities and returns capabilities that should be negotiated back. But once again I suggest to offload users from picky FUSE kernel interaction details with options that has well-defined semantic like e.g PreciseDataCacheControl from the above. And if this way we can't serve a particular use-case - analyze that use case to see what would be the best approach to move forward.

Please let me know what you think.

Thanks beforehand,
Kirill

hanwen: in this case, if the bool is setting just one capability bit, it makes more sense to call the bool

  DisableCapAutoInvalData

and document what it does. This gives us an obvious pattern to follow if we want to expose more capability bits.

I agree that offering the capability flags directly is probably not useful

navytux: Thanks for feedback.

> DisableCapAutoInvalData

It is how it was initially sitting in my dirty tree. However when actually working on this I converged to the idea that the option should negotiate to kernel that _"I, the filesystem, is careful to correctly propagate data invalidation from the backed. You, the kernel, please take advantage of it as much as you can"._ From this point of view we should not name the option 1-1 with currently disabled flag, because in the future kernels the option will be not only disabling CAP_AUTO_INVAL_DATA, but also enabling CAP_PRECISE_INVAL_DATA (https://github.com/hanwen/go-fuse/pull/273/commits/0f8bfb89), and maybe doing something more to take full advantage of the filesystem promise.

With this explanation do you think we could keep it this way, or you prefer we expose to go-fuse users internal details of which capabilities to clear / set so that the kernel could be working efficiently?

I suggest we try to keep it in simple way with options having well-defined documented semantic instead of being attached to particular fixed peer behaviour.

hanwen: shall we postpone this decision until we know where the upstream patch is going?

navytux: It could be both ways. It could indeed make sense to wait for upstream feedback and take it into account. However I'm having trouble to get a reply from upstream for ~3 weeks already even on the patches for which the direction was preliminary explicitly agreed: https://lwn.net/ml/linux-fsdevel/cover.1552558717.git.kirr%40nexedi.com/. I'm afraid just waiting for reply on the patch about cache invalidation could turn out to be even longer.

On the other hand, if we are talking talking about go-fuse -- filesystem server interaction, we can already establish the option for filesystem to negotiate to go-fuse library that the filesystem is careful to propagate all data invalidation from the backend. And then it is go-fuse job to negotiate that via the most efficient way available with the kernel.

I mean we can already establish the option at go-fuse level with documented semantic. We just can't take full advantage of that semantic until kernel has corresponding support, but that could be improving incrementally and independently of filesystem server.

I would suggest waiting for kernel feedback till next week, and in case we don't have it by then, land the go-fuse option into master without waiting for kernel indefinitely.

hanwen: OK

is there anything I can do? (Are there people with @google.com addresses on that list?) Or are we waiting for Miklos?

navytux: Yes, we are primarily waiting for Miklos.

Although they are not on the Cc list, here are the people from Google that touched kernel FUSE at least somhow in recent times:

---- 8< ----
2018:

Jann Horn <jannh@google.com>
Kees Cook <keescook@chromium.org>       +2017

2017:

Eric Biggers <ebiggers@google.com>
Mateusz Jurczyk <mjurczyk@google.com>
Matthew Garrett <mjg59@google.com>

2016:

Tahsin Erdogan <tahsin@google.com>
---- 8< ----

Maybe expressing interest on the mailing list and talking to those people too would make sense...

Thanks for trying to help and appologize for not Cc'ing you initially,
Kirill

navytux: Maybe Andrew Morton <akpm@linux-foundation.org> should be able to help the most (if I understand correctly he is from your company too)

navytux: Still no reply from Miklos. Kernel patches resent for the second time:

https://lwn.net/ml/linux-fsdevel/cover.1553677194.git.kirr%40nexedi.com/

navytux:
( rebased on top of current master )

fuse/api.go
hanwen: looks like nobody that's really involved with FUSE :-/

navytux: indeed...

navytux: @hanwen, there is no reply to kernel FUSE patches at all. Should we merge this as is now?

navytux: @hanwen, the kernel patch has landed into fuse.git#for-next:

https://git.kernel.org/pub/scm/linux/kernel/git/mszeredi/fuse.git/commit/?id=ad2ba64dd489

navytux:
rebased on top of master with "precise" -> "explicit" rename and adjusted commit messages. Interdiff:

```diff
diff --git a/fuse/api.go b/fuse/api.go
index ae5d746..9498065 100644
--- a/fuse/api.go
+++ b/fuse/api.go
@@ -158,12 +158,12 @@ type MountOptions struct {
        // If set, ask kernel not to do automatic data cache invalidation.
        // The filesystem is fully responsible for invalidating data cache.
        //
-       // XXX for Linux PreciseDataCacheControl currently disables data cache
+       // XXX for Linux ExplicitDataCacheControl currently disables data cache
        // to be automatically invalidated only on file mtime change. If file size
        // changes, Linux currently unconditionally discards whole cache of the
        // file. See https://github.com/hanwen/go-fuse/pull/273 for kernel +
        // go-fuse patches with corresponding fixes.
-       PreciseDataCacheControl bool
+       ExplicitDataCacheControl bool
 }
 
 // RawFileSystem is an interface close to the FUSE wire protocol.
diff --git a/fuse/opcode.go b/fuse/opcode.go
index 37f2014..db0b4b2 100644
--- a/fuse/opcode.go
+++ b/fuse/opcode.go
@@ -97,7 +97,7 @@ func doInit(server *Server, req *request) {
 
 
        dataCacheMode := input.Flags & CAP_AUTO_INVAL_DATA
-       if server.opts.PreciseDataCacheControl {
+       if server.opts.ExplicitDataCacheControl {
                // XXX this only disables automatic invalidations on mtime
                // change, but not on size change.
                //
diff --git a/fuse/test/cache_test.go b/fuse/test/cache_test.go
index 4448d61..ee20746 100644
--- a/fuse/test/cache_test.go
+++ b/fuse/test/cache_test.go
@@ -48,7 +48,7 @@ func setupCacheTest(t *testing.T) (string, *pathfs.PathNodeFs, func()) {
 
        mntOpts := &fuse.MountOptions{
                // ask kernel not to invalidate file data automatically
-               PreciseDataCacheControl: true,
+               ExplicitDataCacheControl: true,
 
                Debug: testutil.VerboseTest(),
        }
```

hanwen:
this was merged as part of your other PR, right?

navytux:
Yes, this was merged as part of https://github.com/hanwen/go-fuse/pull/273 which depended on commits from here by PR.

PR #280, opened 2019-03-29
Nodefs api: inode embedding

hanwen:
@navytux  - I further tuned the API. See https://github.com/hanwen/go-fuse/compare/nodefs-api?expand=1#diff-bbbdb1921eef6f0b707b18b19120e84f

what do you think?


hanwen:
@rfjakob

rfjakob:
So I was a little concerned about a possible performance impact that the interface assertions and dynamic dispatch may have. However, a simple benchmark against loopback (https://github.com/hanwen/go-fuse/pull/282/commits/8454f8eeba910e8cef4bf3a03f3dd81a706d2cbf) showed that the new loopback is slightly *faster*.  The simple benchmark is
```
dd if=/dev/zero of=$MNT/zero bs=1 count=200000
```
with the stupidly tiny blocksize of 1 byte chosen to bring out the call overhead. Anyway, we are faster:
```
old: WRITE: 200000 bytes (200 kB, 195 KiB) copied, 6.09325 s, 32.8 kB/s
new: WRITE: 200000 bytes (200 kB, 195 KiB) copied, 5.6114 s, 35.6 kB/s
```

navytux:
@hanwen, thanks for heads up. I will try to review the new nodefs API and embedding in some time. Hopefully not very far from now...

PR #290, opened 2019-04-08
fixup! rename NodeAttr to StableAttr

navytux:
I saw build error when running tests:

	(neo) (z-dev) (g.env) kirr@deco:~/src/neo/src/github.com/hanwen/go-fuse$ go test ./...
	# github.com/hanwen/go-fuse/nodefs
	nodefs/inode.go:130:30: undefined: NodeAttr
	FAIL    github.com/hanwen/go-fuse/benchmark [build failed]
	...

### DIFF
--- a/nodefs/inode.go
+++ b/nodefs/inode.go
@@ -127,7 +127,7 @@ func (n *Inode) setEntryOut(out *fuse.EntryOut) {
 }
 
 // StableAttr returns the (Ino, Gen) tuple for this node.
-func (n *Inode) StableAttr() NodeAttr {
+func (n *Inode) StableAttr() StableAttr {
 	return n.nodeAttr
 }
 


nodefs/inode.go
```diff
@@ -127,7 +127,7 @@ func (n *Inode) setEntryOut(out *fuse.EntryOut) {
 }
 
 // StableAttr returns the (Ino, Gen) tuple for this node.
-func (n *Inode) StableAttr() NodeAttr {
+func (n *Inode) StableAttr() StableAttr {
```
hanwen: aha, perl -i~ -pe s/a/b/ only does one substitution per line :-/

hanwen:
d'oh. I did some more renaming. And tested before pushing. :)

nodefs/inode.go
navytux: :)

navytux:
Thanks.

navytux:
> And tested before pushing. :)

Always a good idea.

PR #293, opened 2019-04-08
tests: gate FUSE debug output behind DEBUG=1

rfjakob:
Running "go test -v" is somewhat verbose as it prints
about 25000 lines.

This patch only prints the FUSE debug output when
DEBUG=1 is set.

Get all 25000 lines of debug logging:

	DEBUG=1 go test ./... -v

Only get individual test results and execution times (476 lines):

	go test ./... -v


hanwen:
What is the use of the -v output without debug info? 

I think we should just stop the giant tests (eg. Stress test) from dumping debug output.

hanwen:
see https://github.com/hanwen/go-fuse/commit/1bba2c11a96f574305d45b9727c70a4aef9db2f2

rfjakob:
Use of -v output is to see the time each test takes - but 1k lines is reasonable, thanks!

rfjakob:
I think some tests were not actually executed when you counted the 1000 lines. We are still at 15k:
```
0 jakob@brikett:~/go/src/github.com/hanwen/go-fuse$ go test -v -count 1 ./... | wc -l
15534
```

Issue #287, opened 2019-04-07
Ubuntu 16.04: test.go:335: ReadDir: readdirent: input/output error


rfjakob:
More trouble with that Ubuntu 16.04 Linux 4.15 kernel.  Fails at https://github.com/hanwen/go-fuse/blob/4f40cf4ccb2e0c8ffa65250f61c7930acce91240/posixtest/test.go#L335 .
```
$ uname -a
Linux osboxes 4.15.0-45-generic #48~16.04.1-Ubuntu SMP Tue Jan 29 18:03:48 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

$ git describe
v1.0.0-170-g4f40cf4

$ go test -run TestReadDir
--- FAIL: TestReadDir (0.07s)
    test.go:335: ReadDir: readdirent: input/output error
FAIL
exit status 1
FAIL	github.com/hanwen/go-fuse/nodefs	0.071s
```
Verbose test log: https://gist.github.com/rfjakob/98702541e427192819b9b997c9807805

hanwen:
could you do an strace to isolate the failing syscall?

rfjakob:
Done:
```
osboxes@osboxes:~$ sudo strace -p 2829
strace: Process 2829 attached
futex(0x7bcca8, FUTEX_WAIT_PRIVATE, 0, NULL) = 0
getdents64(5, /* 21 entries */, 8192)   = 1264
getdents64(5, /* 21 entries */, 8192)   = 1344
getdents64(5, /* 21 entries */, 8192)   = 1344
getdents64(5, /* 20 entries */, 8192)   = 1416
getdents64(5, /* 20 entries */, 8192)   = 1440
getdents64(5, 0xc000198000, 8192)       = -1 EIO (Input/output error)
exit_group(1)                           = ?
+++ exited with 1 +++
```

hanwen:
is this  still a problem now?

rfjakob:
Still the same, seems to be a different issue.
```
osboxes@osboxes:~/go/src/github.com/hanwen/go-fuse/nodefs$ git describe
v1.0.0-173-g7572e9d

osboxes@osboxes:~/go/src/github.com/hanwen/go-fuse/nodefs$ go test
--- FAIL: TestReadDir (0.09s)
    test.go:335: ReadDir: readdirent: input/output error
--- FAIL: TestReadDirStress (0.10s)
    test.go:335: ReadDir: readdirent: input/output error
FAIL
[...]

PR #294, opened 2019-04-14
dirstream_linux: fix zero terminator search

rfjakob:
The old code searched for the first non-null byte from the end of
the slice. This assumes that all bytes after the name are initialized
to zero, which does not hold true on Linux 4.15.

Instead, search for the first null byte from the start of the slice,
which is guaranteed by man(3) readdir.

Fixes https://github.com/hanwen/go-fuse/issues/287


rfjakob:
@hanwen what do you think about having an exported ParseDirent function? gocryptfs also has one ( https://github.com/rfjakob/gocryptfs/blob/cc0a603ef8a691b01fbd0f6a99e7d4e1d36608f7/internal/syscallcompat/getdents_linux.go#L69 ) so there is some potential for code sharing. It's somewhat battle-tested on Linux and MacOS and it has tests, which I would port over.

rfjakob:
Maybe, not sure yet. In any case, I'll try to add some tests.

Issue #298, opened 2019-05-13
SIGSEGV from pathInode.GetAttr() on ESTALE


charles-dyfis-net:
`ENOSYS` and `EBADF` are both handled in the pathfs `GetAttr` implementation, but the Linux-kernel NFS server can return `ESTALE` as well.

This would explain `fi` not being set, and causing a nil pointer exception on line 619 below:

https://github.com/hanwen/go-fuse/blob/509d146ffec41f1d69c7b9ee6cba69e52a9da1a9/fuse/pathfs/pathfs.go#L606-L619

PR #300, opened 2019-05-26
Avoid nil ptr panic in GetAttr()

folbricht:
This PR removes the chance of the nil ptr panic in https://github.com/hanwen/go-fuse/issues/298. Before this change, the only way to avoid a panic is for the `if file == nil || code == fuse.ENOSYS || code == fuse.EBADF` condition to be true, since only then would `fi` be set to something non-nil. So it's not really a condition anymore, and https://github.com/hanwen/go-fuse/commit/47211c2bf0c4ff9da52f38c039ca6538dc6f363d has made it obsolete anyway.


hanwen:
the `fuse/pathfs` library is deprecated.

PR #303, opened 2019-06-23
Renameat2 has been removed and used to implement Renameat instead

jseely:
Upstream change to `golang.org/x/sys/unix` moves the implementation of `Renameat2` to `Renameat` and removes the necessity of `RENAME_EXCHANGE` flag.

### DIFF
--- a/fs/loopback_linux.go
+++ b/fs/loopback_linux.go
@@ -64,7 +64,7 @@ func (n *loopbackNode) renameExchange(name string, newparent *loopbackNode, newN
 		return syscall.EBUSY
 	}
 
-	return ToErrno(unix.Renameat2(fd1, name, fd2, newName, unix.RENAME_EXCHANGE))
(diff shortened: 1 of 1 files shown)


hanwen:
can you point to the commit that introduced this? I am surprised this would work correctly, because RENAME_EXCHANGE is not the default for renameat. Have you tested this?

hanwen:
I think you are referring to https://go.googlesource.com/sys/+/e5ecc2a6747ce8d4af18ed98b3de5ae30eb3a5bb
which uses renameat2 to implement renameat (the other way around). It only really applies to linux/riscv64 since that platform renames the latter.

hanwen:
please reopen with further background if I got confused here

jseely:
Hmm yeah seems you're right. For some reason it wasn't building on my Arch linux system initially and stating that `Renameat2` and `RENAME_EXCHANGE` were missing

Issue #312, opened 2019-08-17
examples don't work due to missing /v2/ import


GregorioDiStefano:
These examples don't work due to this change: 
https://github.com/hanwen/go-fuse/commit/e9498b6bbccf0f17ddd56006dbf8a87a159e8304

The /v2/ folder doesn't exist on master.

Thanks

hanwen:
which version of Go are you using? The v2 paths are for modules.

ShayBox:
Yeah the v2 directory doesn't exist anymore, the examples should be updated, removing the /v2 from the imports fixes the problem.
Module or not it doesn't exist.

hanwen:
Can you confirm this works _with_ modules?

Per https://blog.golang.org/migrating-to-go-modules, 

```
Another scenario in which a module's canonical path may differ from its repository path occurs for Go modules at major version 2 or higher. A Go module with a major version above 1 must include a major-version suffix in its module path: for example, version v2.0.0 must have the suffix /v2. However, import statements may have referred to the packages within the module without that suffix. For example, non-module users of github.com/russross/blackfriday/v2 at v2.0.1 may have imported it as github.com/russross/blackfriday instead, and will need to update the import path to include the /v2 suffix.
```

the v2 suffix is only necessary for the new API, whcih was introduced a few months ago. The old API should continue to work without v2 imports (as evidenced by the unionfs/ subdirectory).

PR #313, opened 2019-08-29
Revert breaking changes

ShayBox:
Linked to #312 
Revert change of module path to /v2, doesn't align with project structure, breaks non-module projects, and breaks pre-change existing projects.

### DIFF
--- a/benchmark/stat_test.go
+++ b/benchmark/stat_test.go
@@ -17,9 +17,9 @@ import (
 	"testing"
 	"time"
 
-	"github.com/hanwen/go-fuse/v2/fs"
-	"github.com/hanwen/go-fuse/v2/fuse"
-	"github.com/hanwen/go-fuse/v2/internal/testutil"
+	"github.com/hanwen/go-fuse/fs"
+	"github.com/hanwen/go-fuse/fuse"
+	"github.com/hanwen/go-fuse/internal/testutil"
 )
 
 func setupFs(node fs.InodeEmbedder, N int) (string, func()) {
--- a/benchmark/statfs.go
+++ b/benchmark/statfs.go
@@ -10,8 +10,8 @@ import (
 	"strings"
 	"syscall"
 
-	"github.com/hanwen/go-fuse/v2/fs"
-	"github.com/hanwen/go-fuse/v2/fuse"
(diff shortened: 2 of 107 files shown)


rfjakob:
Hmm it's not true that it breaks existing project. gocryptfs works just fine, and has not changed anything: https://travis-ci.org/rfjakob/gocryptfs

ShayBox:
> Hmm it's not true that it breaks existing projects. gocryptfs works just fine, and has not changed anything: [travis-ci.org/rfjakob/gocryptfs](https://travis-ci.org/rfjakob/gocryptfs)

It breaks projects that didn't use go modules, as internal files are now referencing /v2 which doesn't match project structure, so projects that don't use go modules won't work, and existing projects that used go modules now have to change the import path to /v2

hanwen:
it only breaks projects that have been using the new fs API (doc'd here: https://godoc.org/github.com/hanwen/go-fuse/fs), which is new and has been under a lot of change up till recently. Just consider this one more example of an API breakage.

ShayBox:
As you can see https://www.youtube.com/watch?v=_3NueLf5XPc

It's a change that was made for no reason, it doesn't match project structure (module and non-module projects now have to use different import paths)
And it's a breaking changeg (module projects now have to update all of their code paths to add /v2)

Even if for some reason you wanted to continue using this breaking change, you would also have to move all of the source into a /v2 folder so the import paths match for module and non module projects, which I doubt you want to do.

hanwen:
The v2 tag was motivated by  https://github.com/hanwen/go-fuse/commit/4dd6878445ae4e5406a6ae6c7629b342226b88ab, which is a breaking change to the API.

I believe my use of the version tags is correct, see for example:

  https://github.com/google/go-github/blob/master/github/examples_test.go

and

  https://github.com/google/go-github/tree/master/#usage

as you can see, there is no v28 directory in the source tree there either. Can you point me to established packages that use a vXXX directories?

If you don't want to use the v2 import, you can keep using the v1 import.

ghz-gehtz:
Hi @ShayBox,

I second @hanwen. I updated my own driver based on go-fuse with the new api (from pathfs to the new fs api) and expect API changes in functions, I didn't have any issue.

I'm using it in a non-module project.

Issue #314, opened 2019-09-09
Allow not depending on fusermount when running as root


MagicalTux:
The fusermount helper allows programs running as users to access the fuse subsystem while performing a number of safety checks (see util/fusermount.c in fuse source), however when the program runs as root it should be possible to avoid needing it at all by directly calling the mount syscall.

At minimum the process on Linux looks like this:
* Open /dev/fuse with O_RDWR
* Call mount syscall, pass fd to opened /dev/fuse in mount option fd=

It feels like this should be easy enough to implement and would remove the need to use fusermount which can be useful in some cases, such as when the executable runs in a read-only embedded system.

hanwen:
see https://review.gerrithub.io/c/hanwen/go-fuse/+/467837 for the review.

hanwen:
https://review.gerrithub.io/c/hanwen/go-fuse/+/467837 was merged.

PR #315, opened 2019-09-09
attempt to mount directly using syscall if running as root

MagicalTux:
Resolves https://github.com/hanwen/go-fuse/issues/314

When attempting to mount fuse, if running as root we first attempt to open /dev/fuse and use syscall.Mount. If either part of the process fails this falls back on the standard process and should typically have no adverse effects.

This is however missing handling of some specific mount options usually handled by fusermount such as blkdev, auto_unmount. In this kind of case syscall.Mount will return EINVAL and the process will fallback to fusermount.

Suggested improvements:
* Add an option to MountOptions to enable this instead of using current effective user id
* Allow customization of mount flags (also MountOptions)

### DIFF
--- a/fuse/mount_linux.go
+++ b/fuse/mount_linux.go
@@ -116,6 +116,13 @@ func mount(mountPoint string, opts *MountOptions, ready chan<- error) (fd int, e
 }
 
 func unmount(mountPoint string) (err error) {
+	if os.Geteuid() == 0 {
+		err := syscall.Unmount(mountPoint, 0)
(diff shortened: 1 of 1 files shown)


hanwen:
thanks!  I usually don't process pull-requests. Can you use the procedure in CONTRIBUTING instead?

MagicalTux:
Submitted to gerrithub:

https://review.gerrithub.io/c/hanwen/go-fuse/+/467837

PR #321, opened 2019-09-29
zipfs: fix "got block count 8, want 1" test failure

rfjakob:
Failure was:

    zipfs_test.go:74: got block count 8, want 1

This got broken by "fs: set Blksize and Blocks automatically based on Size",
fix it by setting Blksize.


hanwen:
thanks, fixed in https://github.com/hanwen/go-fuse/commit/f2ddeec40169901a1cb121599adf02df9aa166a6

PR #343, opened 2019-12-26
FYI: nodefs: Propagate context to File methods

navytux:
[ I know this won't probably be accepted because it has to change nodefs API and nodefs is deprecated. I decided to share anyway since this is general functionality that might be useful ]

So that e.g. File.Read could be canceled.

This is already so in fs/ package. However WCFS in Wendelin.core still
uses older nodefs package, and for now it was faster to teach old
nodefs.File about cancel propagation.

References:

https://lab.nexedi.com/kirr/wendelin.core/commit/b17aeb8c
https://lab.nexedi.com/kirr/wendelin.core/commit/f05271b1
https://lab.nexedi.com/kirr/wendelin.core/commit/5ba816da
https://lab.nexedi.com/kirr/go123/commit/7ad867a3
https://lab.nexedi.com/kirr/go123/commit/0bdac628
https://lab.nexedi.com/kirr/go123/commit/d2dc6c09

### DIFF
--- a/fuse/nodefs/api.go
+++ b/fuse/nodefs/api.go
@@ -144,34 +144,34 @@ type File interface {
 	// the inner file here.
 	InnerFile() File
 
-	Read(dest []byte, off int64) (fuse.ReadResult, fuse.Status)
-	Write(data []byte, off int64) (written uint32, code fuse.Status)
+	Read(dest []byte, off int64, ctx *fuse.Context) (fuse.ReadResult, fuse.Status)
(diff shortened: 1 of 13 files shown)


navytux:
( go1.10 - go1.12 failures seem to be unrelated (TestParallelDiropsHang) and they were kind of present before on other, even trivial, pull requests - e.g. here: https://travis-ci.org/hanwen/go-fuse/jobs/619504219?utm_medium=notification&utm_source=github_status )

hanwen:
thanks for the contribution. You're right that I can't accept this.

PR #355, opened 2020-04-16
Update LICENSE, remove comment starts

MarkAtwood:
License matching tools can have trouble parsing licenses with comment starts in them.


hanwen:
oh, good point. 

see d537a5a8d65c6552bc3844571f1d8ec76bcac5e4.

thanks for the report!

PR #378, opened 2020-11-20
fix deadlock at RmChild

abitduck:


abitduck:
@googlebot I signed it!

hanwen:
yeah, this looks like a deadlock alright. it would be nice if this had a test that triggered this, though.

Also, we're doing code review on Gerrithub (see CONTRIBUTING), but I can forward your change there if you want.

rfjakob:
I don't see the deadlock. lockme includes n.

rfjakob:
Ok, now i see it ;)

Impressive that this survives xfstests.

hanwen:
I was looking at this again, but I can't see the race anymore, and removing the Lock yields:

```
 go test -race
warning: Inode.Path: n2 is orphaned, replacing segment with ".go-fuse.5577006791947779410/deleted"
fatal error: sync: unlock of unlocked mutex

goroutine 443 [running]:
runtime.throw(0x7371a5, 0x1e)
	/usr/lib/google-golang/src/runtime/panic.go:1122 +0x72 fp=0xc000337a28 sp=0xc0003379f8 pc=0x479a72
sync.throw(0x7371a5, 0x1e)
	/usr/lib/google-golang/src/runtime/panic.go:1106 +0x35 fp=0xc000337a48 sp=0xc000337a28 pc=0x4aecf5
sync.(*Mutex).unlockSlow(0xc00016c0d0, 0xffffffff)
	/usr/lib/google-golang/src/sync/mutex.go:196 +0x11a fp=0xc000337a70 sp=0xc000337a48 pc=0x4ca79a
sync.(*Mutex).Unlock(0xc00016c0d0)
	/usr/lib/google-golang/src/sync/mutex.go:190 +0x7b fp=0xc000337a98 sp=0xc000337a70 pc=0x4ca65b
github.com/hanwen/go-fuse/v2/fs.(*Inode).removeRef(0xc00016c080, 0x1, 0x0, 0xc00016c080)
	/home/hanwen/vc/go-fuse/fs/inode.go:428 +0x445 fp=0xc000337c80 sp=0xc000337a98 pc=0x6845c5
```

abitduck:
> I was looking at this again, but I can't see the race anymore, and removing the Lock yields:
> 
> ```
>  go test -race
> warning: Inode.Path: n2 is orphaned, replacing segment with ".go-fuse.5577006791947779410/deleted"
> fatal error: sync: unlock of unlocked mutex
> 
> goroutine 443 [running]:
> runtime.throw(0x7371a5, 0x1e)
> 	/usr/lib/google-golang/src/runtime/panic.go:1122 +0x72 fp=0xc000337a28 sp=0xc0003379f8 pc=0x479a72
> sync.throw(0x7371a5, 0x1e)
> 	/usr/lib/google-golang/src/runtime/panic.go:1106 +0x35 fp=0xc000337a48 sp=0xc000337a28 pc=0x4aecf5
> sync.(*Mutex).unlockSlow(0xc00016c0d0, 0xffffffff)
> 	/usr/lib/google-golang/src/sync/mutex.go:196 +0x11a fp=0xc000337a70 sp=0xc000337a48 pc=0x4ca79a
> sync.(*Mutex).Unlock(0xc00016c0d0)
> 	/usr/lib/google-golang/src/sync/mutex.go:190 +0x7b fp=0xc000337a98 sp=0xc000337a70 pc=0x4ca65b
> github.com/hanwen/go-fuse/v2/fs.(*Inode).removeRef(0xc00016c080, 0x1, 0x0, 0xc00016c080)
> 	/home/hanwen/vc/go-fuse/fs/inode.go:428 +0x445 fp=0xc000337c80 sp=0xc000337a98 pc=0x6845c5
> ```

@hanwen I guess you test with the wrong code change at line 438 and 439 in func `removeRef` , I commit the code is at line 586 and 586 in func `RmChild `

hanwen:
yes, that's it. sorry for the confusion.

hanwen:
How did you find this problem?

I'm asking because Jacob ran this through some extensive stress testing. That makes me think that the  deadlock cannot be triggered from VFS access through the mount point (maybe the VFS serializes removals?). Did this happen because of parallel RmChild's executed from the Go side?

abitduck:
> How did you find this problem?
> 

I implemented some of the fs.NodeXXXer interfaces, for HTTP to POSIX  read-only access, and mount with option as following:

```golang
fs.Options{
    RememberInodes: true, 
    IgnoreSecurityLabels: true, 
    DisableXAttrs: true
}
```

The fuse works good, and the preformance is great [300+MB/s] at first. Then there have more and more inodes, the memory leaks. So i add auto cleaning feature, and do it with `exec.Command("/usr/bin/rm", path)`, Then i find processes state change to D in some containers. I opened the pprof and check the stack of goroutine when fuse not responding. then i find the code with deadlock.

At last. thanks for your go-fuse, that's really awesome!!!

abitduck:
Do rm with parallel  or not have the same result!  In my production environment, there have 5 ~ 10 new inode per second, and all the inodes are the children of root

hanwen:
thanks for the background. 

If you use RememberInodes, you'll get increasing memory usage; even if the kernel drops inodes (issuing FORGET messages), go-fuse would ignore them.

hanwen:
I tried triggering this through the FUSE side, but failed; over here, the VFS serializes the namespace operations. What kernel version did you see this problem with?

hanwen:
merged as https://github.com/hanwen/go-fuse/commit/6dd72b8a8ad3323d76c46b507a768162139a1248

abitduck:
> I tried triggering this through the FUSE side, but failed; over here, the VFS serializes the namespace operations. What kernel version did you see this problem with?

I think i got the problem because of i changed [_MAX_READERS  ](https://github.com/hanwen/go-fuse/blob/master/fuse/server.go#L241) to 256, if we keep the value to 2,  IO may blocks eachother sometimes

PR #385, opened 2021-01-14
fs: panic on mknod if NodeMknoder isn't implemented

tomyl:
`rawBridge.Mknod` currently panics if `NodeMknoder` isn't implemented. This PR makes it return `ENOTSUP` instead.


tomyl:
Push to https://review.gerrithub.io/c/hanwen/go-fuse/+/509897 instead

PR #396, opened 2021-05-30
Add links to examples

i-ky:
This will allow reader to jump to a respective source file with a single click without a need to search for it in a source tree.


i-ky:
@googlebot I signed it!

i-ky:
Obsolete after daaa0cfa8aabe3fd1a3278d886c5680820f09e67.

PR #397, opened 2021-05-31
RFC: Reduce memory usage by recreating big maps

rfjakob:
Posting here as a PR before going through gerrithub because I'd like to know what you think about this.

When we know a lot of inodes the kernelNodeIds & stableAttrs maps grow as needed.
The kernel will send FORGET under memory pressure (simulated here via `echo 3 | sudo tee /proc/sys/vm/drop_caches`), but the maps still take a lot of memory because of https://github.com/golang/go/issues/20135 .

This is an attempt to mitigate this in response to a [user who runs gocryptfs on a 2 GB ARMv7 device](https://github.com/rfjakob/gocryptfs/issues/569).

Commit message follows:

-------------

```
fs: recreate kernelNodeIds & stableAttrs when they shrink

Maps do not free all memory when elements get deleted
( https://github.com/golang/go/issues/20135 ).

As a workaround, we recreate our two big maps (kernelNodeIds & stableAttrs)
when they have shrunk dramatically (100 x smaller).

Benchmarkung with loopback (go version go1.16.2 linux/amd64) shows:

Before this change:

State              VmRSS (kiB)
-----              -----------
Fresh mount          4000
ls -R 500k files   271100
after drop_cache   336448
wait ~ 3 minutes   101588

After:

State              VmRSS (kiB)
-----              -----------
Fresh mount          4012
ls -R 500k files   271100
after drop_cache    31528

Results for gocryptfs are similar.

Fixes: https://github.com/rfjakob/gocryptfs/issues/569
Change-Id: Idcae1ab953270516735839a034d586717647b8db
```

### DIFF
--- a/fs/bridge.go
+++ b/fs/bridge.go
@@ -7,6 +7,7 @@ package fs
 import (
 	"context"
 	"log"
+	"runtime/debug"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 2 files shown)


hanwen:
sounds sane. Can you compare cap(map) with len(map) to decide when to reallocate?

rfjakob:
Unfortunately, there seems to be no way to find out the capacity of a map ( https://groups.google.com/g/golang-nuts/c/5F_xzrm-rJM )

rfjakob:
https://review.gerrithub.io/c/hanwen/go-fuse/+/518681

PR #402, opened 2021-09-20
Extend DirStream for readdirplus

josh-newman:
Lets DirStream implementers optionally implement Plus themselves. In my use case (S3 filesystem), I want to avoid the additional lookup that the bridge does by default because my implementation of DirStream already has all of the data required for Readdirplus.

### DIFF
--- a/fs/api.go
+++ b/fs/api.go
@@ -377,6 +377,13 @@ type DirStream interface {
 	Close()
 }
 
+// DirPlusStream implements readdirplus without an extra Lookup for each entry.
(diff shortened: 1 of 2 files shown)


josh-newman:
If this change seems (roughly) acceptable, I'm happy to update additional documentation, etc. if something's missing here. Thanks!

hanwen:
I think this is OK in principle. Could you use Gerrithub as described in the CONTRIBUTING doc? I can't stand the github review UI.

PR #412, opened 2022-01-04
fix objectpool leak and exchanging child

hotaery:


hotaery:
> 

Is that right? 
https://review.gerrithub.io/c/hanwen/go-fuse/+/530580/1

hanwen:
yes, thanks.

PR #425, opened 2022-05-14
Make OSXFUSE handle timeouts the same way as Linux

### LINKED ISSUE #331: Timeout handling inconsistent between Linux and macOS
The Linux kernel doesn't seem to apply any I/O timeouts on FUSE operations. For FUSE daemons written in Go, this is generally speaking a pretty decent policy. Timeouts can in most cases be added by making proper use of package "context". This not only ensures that processes get unstuck, it also causes resources to be cleaned up on the daemon's side.

OSXFUSE, however, offers an additional mount option called "daemon_timeout". By default it is set to 60 seconds. When I/Os take longer than 60 seconds, OSXFUSE effectively kills the entire mount (read: not just the single I/O). No special signal and/or message is sent to the FUSE daemon when this happens, nor is any functionality offered to recover the mount. All I/Os on the mount will return ENXIO from that moment on.

This feature is pretty annoying when the FUSE file system is backed by remote resources that may sometimes take a longer time to load (e.g., large files that cannot be loaded in chunks).

Would it make sense for go-fuse to simply pass in "daemon_timeout=0" (infinite timeout) on those systems?

Jille:
OSXFUSE applies a 60-second timeout for file operations. This
is inconsistent with how FUSE works on Linux, where operations
last as long as the daemon is willing to let them run.

This has been running in production for ~2 years.

closes #331

### DIFF
--- a/fuse/server.go
+++ b/fuse/server.go
@@ -262,6 +262,13 @@ func (o *MountOptions) optionsStrings() []string {
 		r = append(r, "subtype="+o.Name)
 	}
 
+	// OSXFUSE applies a 60-second timeout for file operations. This
+	// is inconsistent with how FUSE works on Linux, where operations
+	// last as long as the daemon is willing to let them run.
+	if runtime.GOOS == "darwin" {
+		r = append(r, "daemon_timeout=0")
+	}
+
 	return r
 }
 


hanwen:
could you send me a review through gerrithub? See https://github.com/hanwen/go-fuse/blob/master/CONTRIBUTING

the same for the other PRs.

Jille:
Done.

@SnoozeThis https://review.gerrithub.io/c/hanwen/go-fuse/+/539826

SnoozeThis:
(https://snoozeth.is/aytiWqKkVmw) I will wait until https://review.gerrithub.io/c/539826 is merged and then add a comment.

SnoozeThis:
Resolved.

hanwen:
merged as d01fda7edf17995ff56e9b7d8c3f8114e4ae37ce

PR #457, opened 2023-03-27
Export unmount function for Linux and Darwin

### LINKED ISSUE #455: Export unmount function
The `fuse` package has an unexported function called `unmount`, which would be useful for cleaning up mounts in the event of an ungraceful program exit where `Server.Unmount()` could not be called. Would it be possible to rename the `unmount` function to `Unmount` to make it accessible?

Here's the function I'm talking about:

- Linux: https://github.com/hanwen/go-fuse/blob/144cb541affb5b007eefe3821261ad6fe728336e/fuse/mount_linux.go#L175
- Darwin: https://github.com/hanwen/go-fuse/blob/144cb541affb5b007eefe3821261ad6fe728336e/fuse/mount_darwin.go#L89

Arsen6331:
This PR renames the `unmount` function in the `fuse` package to `Unmount`, in order to make it accessible to users of the package.

Fixes #455

### DIFF
--- a/fuse/mount_darwin.go
+++ b/fuse/mount_darwin.go
@@ -86,7 +86,7 @@ func mount(mountPoint string, opts *MountOptions, ready chan<- error) (fd int, e
 	return fd, err
 }
 
-func unmount(dir string, opts *MountOptions) error {
+func Unmount(dir string, opts *MountOptions) error {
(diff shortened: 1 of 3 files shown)


hanwen:
see https://github.com/hanwen/go-fuse/issues/455

PR #459, opened 2023-04-11
Add option for lazy unmount

itomsawyer:
Refers to http://man.he.net/man1/fusermount

LazyMount allow go-fuse unmount works even if resource is still busy (opened by other user programs).

### DIFF
--- a/fuse/mount_linux.go
+++ b/fuse/mount_linux.go
@@ -191,7 +191,7 @@ func unmount(mountPoint string, opts *MountOptions) (err error) {
 	errBuf := bytes.Buffer{}
 	args := []string{"-u", mountPoint}
 	if opts.LazyUnmount {
-		args = append(args, "-z")
+		args[0] = "-uz"
 	}
 	cmd := exec.Command(bin, args...)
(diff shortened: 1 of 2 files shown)


fuse/mount_linux.go
```diff
@@ -189,7 +189,11 @@ func unmount(mountPoint string, opts *MountOptions) (err error) {
 		return err
 	}
 	errBuf := bytes.Buffer{}
-	cmd := exec.Command(bin, "-u", mountPoint)
+	args := []string{"-u", mountPoint}
+	if opts.LazyUnmount {
+		args = append(args, "-z")
+	}
```
marius-enlock, line 195:
```
	args := []string{"-u", mountPoint}
	if opts.LazyUnmount {
		args[0] = "-uz"
	}
```

Looks better to me as it avoids re-sizing the size

marius-enlock:
I am also interested in this one

fuse/mount_linux.go
itomsawyer: Indeed

PR #467, opened 2023-05-24
fusermount: Fix option escaping

levinericzimmermann:
Hello @hanwen,

for the WCFS project [we need to escape commas in fusermount options](https://lab.nexedi.com/nexedi/wendelin.core/merge_requests/15):

> Before this patch no commas were allowed in any fusermount option string. But it is in fact possible to use commas in fusermount options: we only need to escape them [1].
>
> Now go-fuse users can use commas in fusermount options.
> 
> Related libfuse commits are [1] and [2].
>
> ---
>
> [1] https://github.com/libfuse/libfuse/commit/555d6b504308eac6b976321ce938ee4bec62c354
> [2] https://github.com/libfuse/libfuse/commit/5c094ac0150ebfef6a2c9c2c9d1c545a90ff4e96

In [sshfs commas were only escaped if FUSE_VERSION was >= 27](https://github.com/libfuse/sshfs/commit/be44229c), but [Kirill noted that the necessary changes are already very old and we could assume all nowadays used versions already support escaping](https://lab.nexedi.com/nexedi/wendelin.core/merge_requests/15#note_184090).

When running `bash all.bash` i couldn't see any failed tests, it seems all passed ok.

(I thought for a moment if it would make sense to remove user escaped commas before escaping commas in option strings, something like

```go
strings.Replace(strings.Replace(optionValue, "//,". ","), ",", "//,", -1)
```

to avoid constructs as `"////,"`, but I thought, it's too complicated).

Best, Levin

### DIFF
--- a/fuse/server.go
+++ b/fuse/server.go
@@ -185,12 +185,6 @@ func NewServer(fs RawFileSystem, mountPoint string, opts *MountOptions) (*Server
 		o.Name = strings.Replace(name[:l], ",", ";", -1)
 	}
 
-	for _, s := range o.optionsStrings() {
-		if strings.Contains(s, ",") {
-			return nil, fmt.Errorf("found ',' in option string %q", s)
-		}
-	}
-
 	maxReaders := runtime.GOMAXPROCS(0)
 	if maxReaders < minMaxReaders {
 		maxReaders = minMaxReaders
@@ -251,6 +245,10 @@ func NewServer(fs RawFileSystem, mountPoint string, opts *MountOptions) (*Server
 	return ms, nil
 }
 
+func escapeComma(optionValue string) string {
+	return strings.Replace(optionValue, ",", "\\,", -1)
+}
+
 func (o *MountOptions) optionsStrings() []string {
 	var r []string
 	r = append(r, o.Options...)
@@ -273,7 +271,14 @@ func (o *MountOptions) optionsStrings() []string {
 		r = append(r, "daemon_timeout=0")
 	}
 
-	return r
+	// Commas in an option need to be escaped, because
+	// options are separated by a comma.
+	var rEscaped []string
+	for _, s := range r {
+		rEscaped = append(rEscaped, escapeComma(s))
+	}
+
+	return rEscaped
 }
 
 // DebugData returns internal status information for debugging
--- a/fuse/test/mount_test.go
+++ b/fuse/test/mount_test.go
@@ -220,3 +220,19 @@ func TestLiveness(t *testing.T) {
 		t.Fatalf("ReadDir: %v", err)
 	}
 }
+
+// Test that fusermount doesn't exit when when using commas in options.
+func TestEscapedMountOption(t *testing.T) {
+	opts := &fuse.MountOptions{
+		FsName: "fsname,with,many,commas,",
+	}
+	mnt := t.TempDir()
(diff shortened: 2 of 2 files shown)


levinericzimmermann:
(Initial patch wrongly used slash, [but it needs to be backslash](https://github.com/libfuse/libfuse/blob/5c094ac0150ebfef6a2c9c2c9d1c545a90ff4e96/lib/fuse_opt.c#L127-L128).)

levinzimmermannn:
I amended the patch by adding a test and pushed all of this to gerrit ([1](https://review.gerrithub.io/c/hanwen/go-fuse/+/554640), [2](https://review.gerrithub.io/c/hanwen/go-fuse/+/554639)) with a new github/gerrit account which uses my nexedi based mail, so the first contributors CLA [1] shouldn't be a problem anymore.

[1] x https://github.com/hanwen/go-fuse/commit/2adf5e7eb014f9d92993126c24fc7496d39e011d PR Opener: @levinericzimmermann <levin***********rmann​@posteo.eu>

levinericzimmermann:
merged as https://github.com/hanwen/go-fuse/commit/90b055af4668fce08c961406c32c15bbd7953fe4

thanks @hanwen for review & applying the patch!

rfjakob:
I know I'm late to the party, but how does this work with DirectMount? Does
the kernel also understand this escaping?

levinzimmermannn:
Hello @rfjakob,

I'm not a fuse expert, but quickly checking:

```bash
git grep "optionsStrings"
```

showed that `optionsStrings` is only called twice:

1. In `fuse/mount_linux.go` [in `callFusermount`](https://github.com/hanwen/go-fuse/blob/90b055a/fuse/mount_linux.go#L107)
2. In `fuse/mount_darwin.go` [in `mount`](https://github.com/hanwen/go-fuse/blob/90b055a/fuse/mount_darwin.go#L45).

In `mount_linux.go` the `optionsStrings` func is only relevant if `DirectMount` is false. In other words, for linux the applied patch shouldn't affect `DirectMount`.

In `mount_darwin.go` there doesn't seem to be a difference whether `DirectMount` is set or not. I don't know if `mount_macfuse` or `mount_osxfuse` rather resembles `fusermount` or `mount`?

Regarding whether the kernel understands the escaping: I don't know. Since https://github.com/libfuse/libfuse/commit/555d6b504308eac6b976321ce938ee4bec62c354 `fusermount.c` explicitly handles escape characters in string options ([1](https://github.com/libfuse/libfuse/blob/e0041ad/util/fusermount.c#L798-L799), [2](https://github.com/libfuse/libfuse/blob/e0041ad/util/fusermount.c#L744)). [In a test, added with this patch](https://github.com/hanwen/go-fuse/blob/90b055a/fuse/mount_linux_test.go#L194-L209), the server starts and the `,` and `/` are propagated to /proc/mounts.

Best, Levin

navytux:
Hello everyone.

I think mounting directly with "," and "\" likely should work out of the box, because it was only libfuse which needs to do the escaping to be able to separate fsname from other options passed on the command line to fusermount. But the kernel probably does not need to do any escape/unescape because fsname is passed to mount syscall as dedicated argument.

In my view  the only thing which we need to do to complete "fsname with comma" story is to amend TestDirectMount and add cases with comma and backslash there.

@levinericzimmermann, do you think you could do that?

PR #470, opened 2023-06-08
Fix panic when `/dev/null` does not exist

bduffany:
Fixes an issue that the `init()` func in splice.go panics if /dev is not yet mounted.

This allows supporting go-based init binaries which set up all the filesystems on the machine, including both /dev as well as FUSE-mounted dirs.

### DIFF
--- a/splice/splice.go
+++ b/splice/splice.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"sync"
 	"syscall"
 )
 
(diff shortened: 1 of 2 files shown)


splice/splice.go
```diff
@@ -51,13 +55,17 @@ func init() {
...
 
-	fd, err := syscall.Open("/dev/null", os.O_WRONLY, 0)
-	if err != nil {
-		log.Panicf("splice: %v", err)
+// We empty pipes by splicing to /dev/null.
+func devNullFD() int {
+	devNullFDOnce.Do(func() {
+		devNullFDValue, devNullFDErr = syscall.Open("/dev/null", os.O_WRONLY, 0)
+	})
+	if devNullFDErr != nil {
```
hanwen, line 65: should be in the Do func()

hanwen:
if possible, I'd prefer a gerrithub review.

splice/splice.go
bduffany: Fixed

bduffany:
> if possible, I'd prefer a gerrithub review.

Sorry, I am not super comfortable with the permissions that gerrithub is requesting from my GitHub account, and I don't see a way to log in without a GitHub link :confused:

rfjakob:
This has already been merged as https://github.com/hanwen/go-fuse/commit/b5c46276a19e02b7912f998e4ccee85330458f4d

Issue #473, opened 2023-06-08
TestFopenKeepCache is still flaky


navytux:
( this is follow-up to https://github.com/hanwen/go-fuse/issues/168 )

I noticed the following while running go-fuse tests in a loop:

```
--- FAIL: TestFopenKeepCache (0.13s)
    cache_test.go:153: ReadFile: got "before" after notify, want "afterX"
21:31:56.231546 writer: Write/Writev failed, err: 2=no such file or directory. opcode: RELEASE
FAIL
FAIL    github.com/hanwen/go-fuse/fuse/test     1.763s
```

The issue reproduces reliably on my computer in around 30-300 second by running only this test in a loop:

```console
(neo) (z-dev) (g.env) kirr@deca:~/src/neo/src/github.com/hanwen/go-fuse/fuse/test$ time while true; do go test -run TestFopenKeepCache  -count 100 ./... || break ; done
ok      github.com/hanwen/go-fuse/fuse/test     14.777s
ok      github.com/hanwen/go-fuse/fuse/test     14.354s
ok      github.com/hanwen/go-fuse/fuse/test     14.558s
ok      github.com/hanwen/go-fuse/fuse/test     14.091s
ok      github.com/hanwen/go-fuse/fuse/test     14.583s
ok      github.com/hanwen/go-fuse/fuse/test     14.720s
--- FAIL: TestFopenKeepCache (0.15s)
    cache_test.go:153: ReadFile: got "before" after notify, want "afterX"
FAIL
FAIL    github.com/hanwen/go-fuse/fuse/test     14.848s
FAIL

real    1m44,305s
user    0m9,809s
sys     0m12,145s
(neo) (z-dev) (g.env) kirr@deca:~/src/neo/src/github.com/hanwen/go-fuse/fuse/test$ time while true; do go test -run TestFopenKeepCache  -count 100 ./... || break ; done
ok      github.com/hanwen/go-fuse/fuse/test     14.731s
ok      github.com/hanwen/go-fuse/fuse/test     14.661s
ok      github.com/hanwen/go-fuse/fuse/test     14.660s
ok      github.com/hanwen/go-fuse/fuse/test     14.836s
ok      github.com/hanwen/go-fuse/fuse/test     14.719s
ok      github.com/hanwen/go-fuse/fuse/test     14.678s
ok      github.com/hanwen/go-fuse/fuse/test     14.801s
ok      github.com/hanwen/go-fuse/fuse/test     14.733s
[...]

PR #474, opened 2023-06-10
.github: also test GOMAXPROCS=1

rfjakob:
We already had this on Travis CI ( ffb96eff5caa5c4faffb3d5ec4b8de5e75fd1e30 ) but lost it on the migration to Github Actions.

Relates-to: https://github.com/hanwen/go-fuse/issues/473
Change-Id: I1c987fe565227f707d50ea440929f9915a1e86e3

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,9 @@ jobs:
           - "1.18.x" # Golang upstream stable
           - "1.19.x" # Golang upstream stable
           - "1.20.x" # Golang upstream stable
+        GOMAXPROCS:
+          - ""  # Use all cpus (default).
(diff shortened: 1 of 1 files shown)


navytux:
@rfjakob, thanks for restoring `$GOMAXPROCS=1` into test matrix. I think the patch is ok and we better merge it sooner because of how it caught so many failures just in one run, not only TestFopenKeepCache. Thanks once again for restoring this.

/cc @hanwen

navytux:
The patch was applied in https://github.com/hanwen/go-fuse/commit/238293d891ac198a02326513ec6084156cdcc55b. @rfjakob thanks once again for restoring `GOMAXPROCS=1` into test matrix.

Issue #487, opened 2023-08-30
`Lseek` does not handle `ENXIO`


turan18:
The current implementation of `Lseek` with whence `SEEK_DATA` set always returns the current offset of the file: https://github.com/hanwen/go-fuse/blob/0b3e1fde6eb5eeb1600388535913ee66f23bdc5b/fs/bridge.go#L1176

`SEEK_HOLE` always returns the end of the file:

https://github.com/hanwen/go-fuse/blob/0b3e1fde6eb5eeb1600388535913ee66f23bdc5b/fs/bridge.go#L1186

According to https://man7.org/linux/man-pages/man2/lseek.2.html the filesystem should return an `ENXIO` when the offset is greater than the file size on either  `SEEK_DATA` or `SEEK_HOLE` **or**  if the offset is set to the end of the file with `SEEK_DATA` (where there is a hole). 

Re man page:
```
ENXIO   whence is SEEK_DATA or SEEK_HOLE, and offset is beyond the
        end of the file, or whence is SEEK_DATA and offset is
        within a hole at the end of the file.
```

For some context, we had an OracleLinux container running with `soci-snapshotter`, a lazy loading containerd snapshotter, that creates FUSE filesystems for layers in an image. We had issue where a call to `cp` on a sparse file would hang indefinitely. With some digging we found that the command  was making calls to `Lseek` in an infinite loop after the contents of the file were read and the offset was set to the end of the file. We found that this was because `Lseek(fd, EOF_OFFSET, SEEK_DATA)` would keep returning the offset at the end of the file instead of throwing an `ENXIO`.

I have a working implementation and would be happy to contribute upstream if that's fine.

hanwen:
> I have a working implementation and would be happy to contribute upstream if that's fine.

SGTM. Please see CONTRIBUTING for instructions.

turan18:
Hey, the PR is up: https://github.com/hanwen/go-fuse/pull/488

PR #488, opened 2023-09-05
fs: bridge: Add missing Lseek whence checks

turan18:
The man page for `Lseek` says that the filesystem should return an `ENXIO` when the offset is greater
than the file size on either `SEEK_DATA` or `SEEK_HOLE` or if the offset is set to the end of the file with `SEEK_DATA` (where there is an implicit hole). Adding these checks to the current  `Lseek` implementation.

Resolves: #487


hanwen:
needs a test

hanwen:
since this is generic filesystem behavior, you can add a test under posixtest

hanwen:
see https://review.gerrithub.io/c/hanwen/go-fuse/+/1168743

turan18:
> needs a test

Added some tests under posixtest. PTAL: @hanwen

fuse/types.go
```diff
@@ -35,6 +35,9 @@ const (
 	// EINVAL Invalid argument
 	EINVAL = Status(syscall.EINVAL)
 
+	// ENXIO No such device or address
+	ENXIO = Status(syscall.ENXIO)
```
turan18, line 39: Re: https://github.com/hanwen/go-fuse/commit/3dbe3569afbd4fb9d8d06c76202bb77d9f3248b8#r127107015, should we refer to `syscall.ENXIO` directly? @hanwen

turan18: Sure

turan18:
Hey, could you take another look when you the chance? @hanwen

hanwen:
i'll merge it today. sorry for the delay

hanwen:
https://github.com/hanwen/go-fuse/commit/fc2c4d3da0a07dd49fa7c826f406c8e9c2842f32

PR #517, opened 2024-05-28
feat: export `release_in` params

SuzukiHonoka:
Sometimes we need to access the flags which first used by `Open` or other infomations that may match we needs.
Exporting `release_in` is the way to retrive the infomations.


hanwen:
please see CONTRIBUTING for how to send changes. 

you are right that this ReleaseIn should be passed along. However, your change breaks the API, so we can't simply do this.

For the moment, I suggest to remember the flags as they were passed into Open/Create in your filesystem implementation.

SuzukiHonoka:
Thank you for your response!

Issue #504, opened 2024-03-11
Race in Inode.Forgotten()


iain-macdonald:
I recently discovered a [bug in a go-fuse client](https://github.com/containerd/stargz-snapshotter/issues/1594) related to the [`Inode.Forgotten()`](https://github.com/hanwen/go-fuse/blob/e9e7c22af17af4611b5783a16458647088cc8dec/fs/inode.go#L255) function. Roughly speaking, the client has logic for creating and reusing Inode Inos, and relies on calling `Inode.Forgotten()` to detect when Inos can be reused. However, Forgotten() returns `true` briefly between [Inode construction](https://github.com/hanwen/go-fuse/blob/e9e7c22af17af4611b5783a16458647088cc8dec/fs/inode.go#L354) and [Inode initialization](https://github.com/hanwen/go-fuse/blob/e9e7c22af17af4611b5783a16458647088cc8dec/fs/bridge.go#L209-L210), so races can occur and lead to Ino reuse. This is a bit awkward to fix in the client, so I thought I'd ask if `Inode.Forgotten()` should really return `true` if `Inode.lookupCount == 0` as in that case the Inode has not actually been initialized yet, or alternatively, if an `Initialized()` function could be added allowing the caller to check if an Inode has been initialized or not?

hanwen:
> Inode.Forgotten() to detect when Inos can be reused.

I would have to think this through in depth to double-check, but Forgotten() is intended for enabling cleanup in a scenario when you want to be sure the inode cannot be revived. If you are (re)using the inodes, then clearly you are reviving the inode. In a reuse scenario, the value of `Forgotten()` (or any other call that does no locking) can always be out of date by the time you make a decision based on the value.

If you want to securely reuse inodes, I'd put the eligible inodes in a global queue protected by mutex. You can put them in the queue if they are `Forgotten`. For construction, you can take them out of the queue one by one.

Does that make sense?

iain-macdonald:
[...]

PR #520, opened 2024-06-11
Add inode.Initialized() for checking if an inode has been initialized.

iain-macdonald:
This is the simplest fix I could come up with to address [this bug](https://github.com/hanwen/go-fuse/issues/504) in the `Forgotten()` implementation. Happy to change the behavior of `Forgotten` instead, but I figured this was less invasive.


hanwen:
as discussed in https://github.com/hanwen/go-fuse/issues/504 , I don´t want to go this route.

PR #525, opened 2024-08-08
Use `pipe` to reserve fd 3

tempoz:
`/dev/null` is not always available when `go-fuse`'s `init` function runs, so instead of opening `/dev/null`, we make use of `socketpair` and `dup` to reserve file descriptor 3, as it is necessary to prevent deadlocks.

### DIFF
--- a/fuse/mount.go
+++ b/fuse/mount.go
@@ -7,7 +7,7 @@ import (
 	"syscall"
 )
 
-var reservedFDs []int
+var reservedFDs []*os.File
 
 func init() {
 	// Both Darwin and Linux invoke a subprocess with one
@@ -17,23 +17,24 @@ func init() {
 	// grabbing fd 3 and never releasing it. (This is not
 	// completely foolproof: a preceding init routine could grab fd 3,
 	// and then release it later.)
-	fdPair := []int{0, 0}
+	pipes := []*os.File{nil, nil}
 	for {
-		err := syscall.Pipe2(fdPair, syscall.O_CLOEXEC | syscall.O_NONBLOCK)
+		var err error
+		pipes[0], pipes[1], err = os.Pipe()
 		if err != nil {
-			panic(fmt.Sprintf("pipe2([]int{0, 0}, O_CLOEXEC | O_NONBLOCK): %v", err))
+			panic(fmt.Sprintf("os.Pipe(): %v", err))
 		}
-		if fdPair[0] > 3 || fdPair[1] > 3 {
-			for fd := range(fdPair) {
-				if fd > 3 {
-					syscall.Close(fd)
+		if pipes[0].Fd() > 3 || pipes[1].Fd() > 3 {
+			for _, pipe := range(pipes) {
+				if pipe.Fd() > 3 {
+					pipe.Close()
 				} else {
-					reservedFDs = append(reservedFDs, fd)
+					reservedFDs = append(reservedFDs, pipe)
 				}
 			}
 			break
 		}
-		reservedFDs = append(reservedFDs, fdPair...)
+		reservedFDs = append(reservedFDs, pipes...)
 	}
 }
 
--- a/fuse/mount.go
+++ b/fuse/mount.go
@@ -21,7 +21,7 @@ func init() {
 	for {
 		err := syscall.Pipe2(fdPair, syscall.O_CLOEXEC | syscall.O_NONBLOCK)
 		if err != nil {
-				panic(fmt.Sprintf("pipe2([]int{0, 0}, O_CLOEXEC | O_NONBLOCK): %v", err))
+			panic(fmt.Sprintf("pipe2([]int{0, 0}, O_CLOEXEC | O_NONBLOCK): %v", err))
 		}
 		if fdPair[0] > 3 || fdPair[1] > 3 {
 			for fd := range(fdPair) {
--- a/fuse/mount.go
+++ b/fuse/mount.go
@@ -17,23 +17,23 @@ func init() {
 	// grabbing fd 3 and never releasing it. (This is not
 	// completely foolproof: a preceding init routine could grab fd 3,
 	// and then release it later.)
-	fdPair, err := syscall.Socketpair(syscall.AF_UNIX, syscall.SOCK_STREAM | syscall.SOCK_NONBLOCK, 0)
-	if err != nil {
-			panic(fmt.Sprintf("socketpair(AF_UNIX, SOCK_STREAM | SOCK_NONBLOCK, 0): %v", err))
-	}
-	syscall.Close(fdPair[1])
-	fd := fdPair[0]
-
+	fdPair := []int{0, 0}
 	for {
-		if fd > 3 {
-			syscall.Close(fd)
+		err := syscall.Pipe2(fdPair, syscall.O_CLOEXEC | syscall.O_NONBLOCK)
+		if err != nil {
+				panic(fmt.Sprintf("pipe2([]int{0, 0}, O_CLOEXEC | O_NONBLOCK): %v", err))
+		}
+		if fdPair[0] > 3 || fdPair[1] > 3 {
+			for fd := range(fdPair) {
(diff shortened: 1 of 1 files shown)


tempoz:
(Note that I specifically have run into this when trying to run `go-fuse` with newer versions of firecracker

tempoz:
I mean, I'm not an expert on the internal workings of firecracker, I just know I was getting panics like this:
```
[    1.006786] Run /init as init process?
panic: open("/dev/null"): open /dev/null: no such file or directory?
?
goroutine 1 [running]:?
github.com/hanwen/go-fuse/v2/fuse.init.0()?
        external/com_github_hanwen_go_fuse_v2/fuse/mount.go:23 +0x17f?
[    1.024444] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000200?
[    1.027520] CPU: 0 PID: 153 Comm: init Not tainted 5.15.0 #18?
[    1.029851] Call Trace:?
[    1.030876]  dump_stack_lvl+0x38/0x49?
[    1.032447]  dump_stack+0x10/0x12?
[    1.033838]  panic+0xe8/0x27a?
[    1.035075]  do_exit.cold+0x71/0xa0?
[    1.036487]  do_group_exit+0x36/0xa0?
[    1.038006]  get_signal+0x149/0x7f0?
[    1.039436]  arch_do_signal_or_restart+0x101/0x720?
[    1.041400]  ? pick_next_task_fair+0x1b4/0x360?
[    1.043217]  ? __switch_to+0x13e/0x3a0?
[    1.044732]  ? __switch_to_asm+0x36/0x70?
[    1.046319]  ? __x64_sys_futex+0x73/0x1d0?
[    1.047982]  ? __x64_sys_futex+0x73/0x1d0?
[    1.049650]  exit_to_user_mode_prepare+0xb1/0x120?
[    1.051562]  syscall_exit_to_user_mode+0x21/0x40?
[    1.053441]  do_syscall_64+0x48/0x90?
[    1.054907]  entry_SYSCALL_64_after_hwframe+0x44/0xae?
[    1.056959] RIP: 0033:0x475f43?
[    1.058224] Code: 24 20 c3 cc cc cc cc 48 8b 7c 24 08 8b 74 24 10 8b 54 24 14 4c 8b 54 24 18 4c 8b 44 24 20 44 8b 4c 24 28 b8 ca 00 00 00 0f 05 <89> 44 24 30 c3 cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc cc?
[    1.065603] RSP: 002b:000000c000083e48 EFLAGS: 00000286 ORIG_RAX: 00000000000000ca?
[    1.068646] RAX: fffffffffffffe00 RBX: 0000000000000000 RCX: 0000000000475f43?
[    1.071494] RDX: 0000000000000000 RSI: 0000000000000080 RDI: 000000c000063148?
[    1.074348] RBP: 000000c000083e90 R08: 0000000000000000 R09: 0000000000000000?
[    1.077177] R10: 0000000000000000 R11: 0000000000000286 R12: 000000c000083e38?
[    1.080037] R13: 000000c000063008 R14: 000000c000007180 R15: 001fffffffffffff?
[    1.083048] Kernel Offset: disabled?
[    1.084488] Rebooting in 1 seconds..?
2024-08-08T15:09:07.671466029 [643fa7dd-eaf7-400a-b0a1-b6e152a4b866:main] Vmm is stopping.
2024-08-08T15:09:07.671706869 [643fa7dd-eaf7-400a-b0a1-b6e152a4b866:main] Vmm is stopping.
2024-08-08T15:09:07.712478128 [643fa7dd-eaf7-400a-b0a1-b6e152a4b866:main] Firecracker exiting successfully. exit_code=0
```

Presumably this is because we were including `go-fuse` in our custom `/init` binary, which was running before the kernel had set up `/dev/null` (at least on this version of firecracker), and all `init` functions in `go` run as they are loaded, so `go-fuse` cannot currently be imported by an `/init` binary running under these conditions without causing a kernel panic. I understand that this is likely not a standard use-case, and may not be one you wish to support, but personally I don't see the harm in using `socketpair` and `dup` instead to sidestep this issue entirely, so I am hopeful you will consider my pull request.

Thank you very much for following up on this!

hanwen:
In principle this is ok; i'd propose to use syscall.pipe because it looks simpler. Also, why dup the fd? Just create new fd pairs in a loop.

rfjakob:
> Presumably this is because we were including go-fuse in our custom /init binary

Understood. Cool use-case. Looking at https://unix.stackexchange.com/questions/619589/is-it-necessary-to-mount-devtmpfs-with-etc-fstab , /init can not rely on /dev being mounted already, as you said.

I think os.Pipe() is even better because it sets O_CLOEXEC for you. Otherwise you leak the fds to all subprocesses you may create.

rfjakob:
Looks like go-fuse is missing O_CLOEXEC in a few places. I fixed gocryptfs via https://github.com/rfjakob/gocryptfs/commit/da873087dd7bcc54d96748a1f6e4e8ecf9e265e5 .

hanwen:
> Looks like go-fuse is missing O_CLOEXEC in a few places

tell me more?

tempoz:
> In principle this is ok; i'd propose to use syscall.pipe because it looks simpler. Also, why dup the fd? Just create new fd pairs in a loop.

I chose to use `dup` because, per https://man7.org/linux/man-pages/man2/dup.2.html:
```
> The new file descriptor number is guaranteed to be the lowest-numbered file descriptor that was unused in the calling process.
```

Since I didn't see that language for `socketpair` in its man page, I chose to use `dup`. However, after seeing your susggestions to use `pipe` instead, I decided I should look up the specification instead of the man page, and sure enough, per https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_14:

```
All functions that open one or more file descriptors shall, unless specified otherwise, atomically allocate the lowest numbered available (that is, not already open in the calling process) file descriptor at the time of each allocation.
```

So my concern was unfounded. I'll adjust the PR accordingly. Thanks for the input!

tempoz:
I updated the code to use `pipe` (and updated the PR title to reflect that), let me know if there's anything else I can do to help with this!

rfjakob:
>> Looks like go-fuse is missing O_CLOEXEC in a few places
>
> tell me more?

All open fds are passed to new child processes **unless** they are marked as O_CLOEXEC. This is bad, so Go stdlib sets O_CLOEXEC everywhere (example: [os.Open()](https://github.com/golang/go/blob/0253c0f4acdf9f7a2930609f55d24b334fa7d3c2/src/os/file_unix.go#L279), [os.Pipe()](https://github.com/golang/go/blob/master/src/os/pipe_unix.go#L23-L24)).

These all need O_CLOEXEC:

```
go-fuse$ git grep syscall.Open

fs/dirstream_unix.go:   fd, err := syscall.Open(name, syscall.O_DIRECTORY, 0755)
fs/loopback.go: fd, err := syscall.Open(p, int(flags)|os.O_CREATE, mode)
fs/loopback.go: fd1, err := syscall.Open(n.path(), syscall.O_DIRECTORY, 0)
fs/loopback.go: fd2, err := syscall.Open(p2, syscall.O_DIRECTORY, 0)
[...]
```

rfjakob:
This comment also explains the problem (and what to do if your syscall does not support O_CLOEXEC): https://github.com/golang/go/blob/0253c0f4acdf9f7a2930609f55d24b334fa7d3c2/src/syscall/exec_unix.go#L22

rfjakob:
Why not os.Pipe() ?

tempoz:
Oh, no reason, I'd just been using `syscall` and I guess I kept doing it. Here it is with `os.Pipe`!

hanwen:
see https://review.gerrithub.io/c/hanwen/go-fuse/+/1199616

I rewrote it, but left your name as author. Hope that's ok.

tempoz:
Works for me; thanks again for your time and effort!

PR #531, opened 2024-09-13
fuse: replace deprecated StringBytePtr with BytePtrFromString

Stavrospanakakis:
As mentioned in the Go codebase, the method `syscall.StringBytePtr` is deprecated and has to be replaced with `syscall.BytePtrFromString`.

```go
// StringBytePtr returns a pointer to a NUL-terminated array of bytes.
// If s contains a NUL byte this function panics instead of returning
// an error.
//
// Deprecated: Use [BytePtrFromString] instead.
func StringBytePtr(s string) *byte { return &StringByteSlice(s)[0] }

// BytePtrFromString returns a pointer to a NUL-terminated array of
// bytes containing the text of s. If s contains a NUL byte at any
// location, it returns (nil, [EINVAL]).
func BytePtrFromString(s string) (*byte, error) {
	a, err := ByteSliceFromString(s)
	if err != nil {
		return nil, err
	}
	return &a[0], nil
}
```
https://github.com/golang/go/blob/8a4b439ee6e0f5d41cb6ae30146b901f3c0ad4b7/src/syscall/syscall.go#L57-L73

This pull request replaces the deprecated method.

Change-Id: I31b52356438352829971c6b491d37fe09e088911

### DIFF
--- a/fuse/syscall_darwin.go
+++ b/fuse/syscall_darwin.go
@@ -11,8 +11,16 @@ import (
 )
 
 func getxattr(path string, attr string, dest []byte) (sz int, errno int) {
-	pathBs := syscall.StringBytePtr(path)
-	attrBs := syscall.StringBytePtr(attr)
+	pathBs, err := syscall.BytePtrFromString(path)
+	if err != nil {
+		return 0, int(syscall.EINVAL)
+	}
+
+	attrBs, err := syscall.BytePtrFromString(attr)
+	if err != nil {
+		return 0, int(syscall.EINVAL)
+	}
+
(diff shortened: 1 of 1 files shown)


hanwen:
can you upload to gerrithub? thanks. See CONTRIBUTING.

Stavrospanakakis:
@hanwen Sure, I uploaded the code to gerrithub.
https://review.gerrithub.io/c/hanwen/go-fuse/+/1201253

hanwen:
thanks! merged as 58e0d8301f7da5bc9e12f04dccad78cbeee32854

PR #537, opened 2024-10-21
fix nil fileEntry

liver:
on Create, after addNewChild fe(fileEntry) is nil value

### DIFF
--- a/fs/rawbridge_test.go
+++ b/fs/rawbridge_test.go
@@ -1,48 +1,57 @@
 package fs
 
 import (
+	"context"
+	"os"
+	"syscall"
 	"testing"
+	"time"
+
 	"github.com/hanwen/go-fuse/v2/fuse"
 )
 
-// Create a dummy rawBridge for testing
-type dummyRawBridge struct{}
-
-// Define a FileEntry structure with fh that rawBridge can use in Create
-type FileEntry struct {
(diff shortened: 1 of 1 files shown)


hanwen:
needs a test that repros the problem.

liver:
accessing a property of an object that does not exist.
For NodeWriter interface, in the case of FileHandle is nil.

hanwen:
needs a test that demonstrates the crash.

hanwen:
Your test should create a file system, mount it, and then trigger the crash. I suspect that creating a file in an FS that doesn't use file handles might do the trick.

liver:
Changed the test.
It seems to me that in this case, the test is superfluous.
I have not changed or added any functionality or logic.
Added a check for the existence of an object before accessing its property.

hanwen:
why are you proposing this fix then?

liver:
fix nil pointer dereference
check for existence of object

liver:
Is there any chance of correction?

```
if fe != nil {
    out.Fh = uint64(fe.fh)
}
```

hanwen:
I am still waiting for a test that demonstrates a crash in a filesystem without this fix. If you have a FS that doesn't work, it should be straightforward to distill that into a test.

PR #582, opened 2025-07-30
fs: Add a condition check for PassthroughFd to avoid invalid ioctl system calls

wswsmao:
Under normal conditions, the file descriptor may still be invalid. For example, in certain scenarios, users can control the return value of PassthroughFd to fine-tune whether passthrough mode should be enabled for a specific file. In such cases, it is possible to return immediately, thereby avoiding the overhead of an unnecessary ioctl system call.


hanwen:
>> why can't PassthroughFd() return false in this case?

you did not answer my question.

> `return -1, nil`

the signature of PassthroughFd is `PassthroughFd() (int, bool)`

PR #583, opened 2025-09-03
fuse: add `EnableWritebackCaching` option to control writeback I/O mode

alberto-miranda:
## Description

This PR adds a new `EnableWritebackCaching` option that allows users to set the `CAP_WRITEBACK_CACHE` when mounting the filesystem. This option changes the way in which write requests are processed reach the filesystem server: instead of each write being immediately sent to userspace as one or more WRITE requests, writes will be satisfied directly from the page cache. Since this option may have negative repercussions in networked filesystems, it is disabled by default.

### DIFF
--- a/fuse/api.go
+++ b/fuse/api.go
@@ -332,6 +332,25 @@ type MountOptions struct {
 
 	// Disable splicing from files to the FUSE device.
 	DisableSplice bool
+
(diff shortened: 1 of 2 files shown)


alberto-miranda:
Apologies, this PR was not supposed to be here (yet). Closing.

PR #585, opened 2025-09-10
fs: support RENAME_WHITEOUT in loopback

hunshcn:
This will help making a upper fs for overlayfs


hunshcn:
@hanwen could you please take a look?

PR #586, opened 2025-10-03
Allow to do the lookup for readdirplus without going through the parent node

bitonic:
In cases where the readdir object already contains the result of the stat, this allows for a much less awkward implementation.

### DIFF
--- a/fs/api.go
+++ b/fs/api.go
@@ -703,10 +703,26 @@ type NodeOpendirHandler interface {
 	OpendirHandle(ctx context.Context, flags uint32) (fh FileHandle, fuseFlags uint32, errno syscall.Errno)
 }
 
+type HasDirEntry interface {
+	GetDirEntry(out *fuse.DirEntry)
+}
+
+type DirEntryLookuper interface {
+	Lookup(ctx context.Context, parent *Inode, out *fuse.EntryOut) (*Inode, syscall.Errno)
+}
+
+type SimpleDirEntry struct {
+	DirEntry *fuse.DirEntry
+}
+
+func (sde SimpleDirEntry) GetDirEntry(out *fuse.DirEntry) {
+	*out = *sde.DirEntry
+}
+
 // FileReaddirenter is a directory that supports reading.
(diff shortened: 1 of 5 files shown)


hanwen:
please see CONTRIBUTING for instructions.

from a quick look, this is not backward compatible, so I cannot accept this. Also, your change lacks motivation: why is this a good idea?

bitonic:
Yes, sorry, I should have maybe phrased this as a proposal, since clearly it's not backwards compatible.

The reason it's (I think) a nicer API is that you can tie the lifecycle of the readdir object to the lifecycle of the lookups. Consider the case where you get the stat information as you get the dir entries. If the lookup is on the same object that's emitting the dir entries, you don't need to communicate the lookup info upstream to the object for the directory. Moreover that data will be freed when the readdir is finished.

The concrete case that prompted me to change the api is for the FUSE client of TernFS (a distributed filesystem), where doing things this way is a lot more natural, since directory objects don't cache lookups at all, instead relying on FUSE's `d_revalidate` to cache direntries for the given timeout: https://github.com/XTXMarkets/ternfs/blob/3aff31a026fc7a4336594644df1b70bab9ab8c01/go/ternfuse/ternfuse.go#L840-L1028 .

hanwen:
thanks for the explanation. I came up with this:  https://review.gerrithub.io/c/hanwen/go-fuse/+/1223814

Does this cover your use case? Happy to hear if you have good suggestions for a realistic test case. The linux syscall API doesn't have a readdirplus.

bitonic:
Yes, that works, thank you, although it's slightly less convenient.

Regarding the test: I'd do it by setting up an instrumented server which records whether you're getting calls into the FileLookuper.

bitonic:
Oh and let me know if you'd prefer for me to come up with the test, I can take a look if you want.

bitonic:
I would say that the common case is what I've outlined already (you can get direntries + stat info at once), which is basically the case for readdirplus.

I suppose I'm not clear of what this test you're envisioning is for, is it about testing performance?

PR #591, opened 2025-11-27
fs: defer initial getdents() in loopbackDirStream

ZRHann:
Some callers (e.g. [propfind in x/net/webdav](https://cs.opensource.google/go/x/net/+/master:webdav/prop.go;l=170)) open a directory only to get a handle for stat, without reading entries. 

Avoiding `getdents()` in `NewLoopbackDirStreamFd()` saves significant latency, especially for remote filesystems. In my case, per-opendir cost drops from ~20ms to <1ms. With hundreds of such opendir calls during a typical WebDAV PROPFIND, the total latency reduction is dramatic.

### DIFF
--- a/fs/dirstream.go
+++ b/fs/dirstream.go
@@ -120,6 +120,7 @@ type loopbackDirStream struct {
 	todo      []byte
 	todoErrno syscall.Errno
 	fd        int
+	inited    bool
 }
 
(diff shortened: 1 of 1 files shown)


ZRHann:
@hanwen When will the change be merged on GitHub? I saw it has already been approved in Gerrit. Thanks!

hanwen:
thanks for the ping, the gerrithub -> github mirroring is not the most reliable.

PR #596, opened 2026-01-11
fs: NewNodeFS: actually handle opts = nil

rfjakob:
Cross-post of https://review.gerrithub.io/c/hanwen/go-fuse/+/1228929 for CI

-----------------

Parts of the function did handle opts = nil, other parts did not.

Handle it right at entry like fuse.NewServer,
and document it.

Change-Id: I265f1d7cb96a926879ca6769a0d08e613795b9cd

### DIFF
--- a/example/loopback/main.go
+++ b/example/loopback/main.go
@@ -16,7 +16,6 @@ import (
 	"path"
 	"runtime/pprof"
 	"syscall"
-	"time"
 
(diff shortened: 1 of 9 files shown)


rfjakob:
Merged as https://github.com/hanwen/go-fuse/commit/9b1593c07a19f8876b354704273d70a7de8c4274

